PR #108, opened 2014-08-19
Reuse function in LayerDraggable.TouchMove

msfeldstein:
This might be an unnecessary optimization but it seems like it could hurt performance to create a new anonymous function in every frame of TouchMove.


koenbok:
This got fixed with the scroll/draggable update a while ago.

PR #113, opened 2014-09-12
LayerStates.switch() returns the animation object

manucorporat:
A use case for this change:
```coffee
animation = layer1.switch("default")
animation.on 'end', ->
    # do some stuff
    otherLayer.visible = false
```


koenbok:
Well why not. Thanks :-)

PR #147, opened 2014-12-23
make icon link work on case-sensitive file systems

Kilian:
Linux (often) uses a case insensitive file system. Linking to "icon.png" but having an "Icon.png" results in a 404. Mac and Windows both use case-insensitive file systems so you don't notice it there.


koenbok:
Thanks!

PR #149, opened 2015-01-06
Switching from absolute to relative dragging

paulmand3l:
This allows a developer to update speedX or speedY on the fly if, for example, they wanted to implement iOS overscrolling.


koenbok:
Worth still looking into...

nvh:
This is interesting, but it definitely needs tests. Are you still interested in merging this @paulmand3l?

PR #164, opened 2015-01-19
Improved scrolling and dragging with physics simulations

dabbott:
Physics-based scrolling. Here we go!

Demos here:

* [Draggable rows in a scrollview](https://db.tt/dYLF2AmP)
* [Crop a photo](https://db.tt/XlOGXQvL)
* [Fling an object](https://db.tt/bJqVil5A)

Demo source [here](https://github.com/dabbott/FramerScrollPhysicsDemos)

tl;dr: It's a hard problem. This is a first draft that works decently, but there are a lot of other ways we could implement this, and it really comes down to Framer's direction.

#### My priorities for this first pass:

1. Scrolling should feel 100% accurate
2. It should be simple to use
3. The math should be straightforward enough that anyone can read and maintain the code
4. Modify existing stuff as little as possible. Don't break Framer

Given that the new [priority list](https://github.com/koenbok/Framer/wiki/Contributing:-Wanted-Features) includes event handling and a physics engine, I could imagine waiting on those two to improve scrolling, as they're really the building blocks. Nonetheless, if we do decide to build some of these components ourselves, I outline some of the challenges and possible solutions below.

#### 1. Scrolling should feel 100% accurate

However we go about it, scrolling has to feel perfect. It's one of the keys to making mobile prototypes feel native.

My first attempt was something along the lines of [ScrollViewBest](http://cl.ly/083x3h3Y0E15), using the existing Animators. The reasons why this doesn't work: 

* for the spring, the "scale" of the motion is determined by the initial and end positions stored in `Animation._stateA` and `Animation._stateB`. By that I mean, igit stf you start a spring animation with `velocity = 100` and final position `y = 0`, the result will be vastly different depending on whether the view starts at postiion `y = 5` versus `y = 5000`.

* an animation from 0 to 1 is really not the easiest way to represent friction. You could have a FrictionAnimator, but you would have to determine the end position of the view completely independently from the actual animation. You would also have to make sure either velocity or friction was set automatically, or there's a chance the animator would stop short.

You could solve this by modifying Animation. Letting animators do more than just animate from 0 to 1 should solve these problems (an unbounded animation for friction, 0 to N for spring). However, then you would have a variety of different behaviors that Animation would have to deal with, which isn't the intended usage of the existing class.

Instead, I solved this by adding a concept of physical Simulators. These are really just a generalized version of Animators. I avoided touching Animators for now (goal 4, don't break Framer), but building Animators on top of a more general physics engine should be fairly easy.

The physics credit really goes to Ralph at [iamralpht.github.io/physics/](iamralpht.github.io/physics/), as I pretty much just implement the Spring + Friction as described. The main difference - I used numerical integration instead of symbolic integration, since the math is much more straightforward (goal 3, simple math). The only place I include any interesting math is the `IntertialScrollSimulator._isValidState()` method, and I provide a step-by-step explanation, which can be checked for correctness.

#### 2. It should be simple to use

I could imagine having improved scrolling either be part of Layer or LayerDraggable (or both):

* Layer has `scroll`, `scrollHorizontal`, `scrollVertical`, and `scrollFrame`. As a Framer user, I would probably assume these are what I want to use. It takes a reasonably good understanding of what these really do and what their limitations are to understand when I should use LayerDraggable instead.

* LayerDraggable has touch event handling and velocity calculation. I like how easy to is to apply a "draggable" interaction to any layer. I could imagine doing "pinchable" or "rotatable" in a similar way. I like that if inertial scroll is part of LayerDraggable, it is more general, and I can use it for interactions other than scrolling e.g. flinging an object. However, *because* "dragging" is much more general than "scrolling", I hesitate to couple something somewhat view-hierarchy-dependent like a "scrollFrame" with LayerDraggable. And I don't want to introduce yet another scrolling concept with `LayerScrollable extends LayerDraggable`.

* Maybe there's an even better option?

Ultimately I went with LayerDraggable. I don't enforce a parent/sublayer relationship, so any view is "scrollable" within any other view (which is actually quite convenient...). 

The main properties:

* `draggable.inertialScroll` - the layer keeps moving after you let go. This starts the physics simulation. `draggable.speedX|Y` work as expected, and can be used to prevent dragging in a direction.
* `draggable.overscroll` - the layer drags with resistance past the edges, and will spring back if inertialScroll is true. Get a more Androidy feel by turning inertialScroll on by overscroll off.
* `draggable.directionLock` - scrolling can lock in a direction.. same as iOS. The implementation: the drag must move `DragThreshold = 10px` before the layer begins to move. Locking can occur in X or Y, or both if the drag moves more than `DragThreshold / 2 = 5px` in both directions.
* `draggable.multipleDraggables` - propagate touch events until directionLock (very hacky)

I didn't want to break anything, so I kept `maxDragFrame` and added a new, equally confusing concept `maxScrollFrame`, which is a helper to set up your `maxDragFrame`. `maxDragFrame` works acceptably for dragging a small layer within a large one, but it's super hard to figure out the proper `maxDragFrame` for scrolling a large layer within a smaller one. While these are two distinct use cases, they're both confusing, and I'd like to merge them somehow.

`directionLock` is a bit of an afterthought to be honest, and should *really* go wherever we put event handling stuff. The DragThreshold should probably scale by deviceScale, and may need to take the velocity of the view into account (since we generally swipe in arcs that go quite a bit horizontal and will trigger a false positive). As a user, I would also wonder why I can't use `directionLock` on Layer itself if I use `scroll`.

The `multipleDraggables` thing is an afterthought tacked on to the `directionLock` afterthought. Having to manually turn on/off certain draggables is pretty terrible and I'm sure we can do much better. Perhaps a single layer in a group can obtain ownership over events for that group, or similar.

Other stuff I'd want to improve:

* There are too many physics constants being passed around
* Defaults for iOS and Android
* Tests

#### 3. The math should be straightforward enough that anyone can read and maintain the code

The reasons I would support writing the physics code ourselves, or perhaps making physics a plugin, versus integrating an existing library into Framer core:

* Physics libraries are big and include a lot of complicated stuff that we don't need. Framer's existing dependencies (lodash, EventEmitter3) are not at all like that. Lodash is a defacto standard that doesn't require additional learning. EventEmitter3 has roughly the standard EventEmitter API and is only ~200 lines.

* Physics libraries are built for high performance with millions of objects in the simulation. Which totally isn't the Framer use case. So the code is *generally* more convoluted and we wouldn't even get the benefit of it. (I don't know how frequently developers/users would have to wade through physics engine code... maybe we're not concerned about this point)

* Part of Framer's beauty is that the codebase is so small and written with clarity as a key goal. The current event handling and animators in Framer are written in a clear, consistent style and super easy to understand. Building a tiny physics engine ourselves lets us keep that consistent style where it really matters.

But I'm biased since I already wrote some code :) I also don't have much experience using the various .js-based physics engines, so I can't speak too much to their complexity.

Anyway, I pulled out the RK4 integrator we already have into a separate class that's shared between the spring Animator and spring/friction Simulators.

My code isn't optimized for performance in any way... lots of unnecessary object creation that would be easy to get rid of without hurting clarity much. Would also need to look into why the simulation blows up under stress.

#### 4. Modify existing stuff as little as possible. Don't break Framer

(minor) existing things I changed:

* I changed how velocity is calculated - based on the average dragging speed over the last 100ms instead of between the last two events. In the end I'm not sure it makes much of a difference. This feels more accurate to me, but maybe that's just me. I could probably acheive the same feel with the previous method and slightly different constants. (also there's a [minor bug in the existing code](https://github.com/koenbok/Framer/blob/e5ce332acba6fe2436c33c7f2b48c95559671a11/framer/LayerDraggable.coffee#L68), as @VelocityTimeOut does not refer to the constructor here)

* I added an `animatingProperties` method to Animation, so that Animations and Simulations can play nice in a layer's `animationList`. The right way for these two to play nice everywhere would probably be a BaseAnimation class which Animations and Simulations extend, or merge Animations and physics Simulations as mentioned above.

* SpringRK4Animator now uses the Integrator class. If we keep this spring class, is could be cleaned up a ton - or just wrap an instance of SpringSimulator with `@_offset = -1`.

#### Next steps

Where do you folks think these pieces fit? If we do actually go with any of what I have here, we would have to improve the API a lot. What changes do you suggest?

Pull in a physics engine engine and get that for free, or write a simpler, more compatible version ourselves?

Pull in an event handling library and get that for free, or write a simpler, more layer-centric version ourselves? Is Framer moving toward WebGL rendering? If so, would we have to maintain a DOM event handling library and our own WebGL event handling library that propagates events via Layers?

Or defer the physics/event library decisions and go with something along these lines for now?

### DIFF
--- a/framer/Animation.coffee
+++ b/framer/Animation.coffee
@@ -169,6 +169,9 @@ class exports.Animation extends EventEmitter
 		# Also emit this to the layer with self as argument
 		@options.layer.emit(event, @)
 
+	animatingProperties: ->
+		_.keys(@_stateA)
+
 	_start: =>
 		@options.layer._context._animationList.push(@)
 		@emit("start")
--- a/framer/Animators/SpringRK4Animator.coffee
+++ b/framer/Animators/SpringRK4Animator.coffee
@@ -1,6 +1,7 @@
 Utils = require "../Utils"
 
 {Animator} = require "../Animator"
+{Integrator} = require "../Integrator"
 
 class exports.SpringRK4Animator extends Animator
 
@@ -18,6 +19,9 @@ class exports.SpringRK4Animator extends Animator
 		@_velocity = @options.velocity
 		@_stopSpring = false
 
+		@_integrator = new Integrator (state) =>
+			return - @options.tension * state.x - @options.friction * state.v
+
 	next: (delta) ->
 
 		if @finished()
@@ -31,11 +35,9 @@ class exports.SpringRK4Animator extends Animator
 		# Calculate previous state
 		stateBefore.x = @_value - 1
 		stateBefore.v = @_velocity
-		stateBefore.tension = @options.tension
-		stateBefore.friction = @options.friction
 		
 		# Calculate new state
-		stateAfter = springIntegrateState stateBefore, delta
+		stateAfter = @_integrator.integrateState stateBefore, delta
 		@_value = 1 + stateAfter.x
 		finalVelocity = stateAfter.v
 		netFloat = stateAfter.x
@@ -52,45 +54,3 @@ class exports.SpringRK4Animator extends Animator
 
 	finished: =>
 		@_stopSpring
-
-
-springAccelerationForState = (state) ->
-	return - state.tension * state.x - state.friction * state.v
-
-springEvaluateState = (initialState) ->
-
-	output = {}
-	output.dx = initialState.v
-	output.dv = springAccelerationForState initialState
-
-	return output
-
-springEvaluateStateWithDerivative = (initialState, dt, derivative) ->
-
-	state = {}
-	state.x = initialState.x + derivative.dx * dt
-	state.v = initialState.v + derivative.dv * dt
-	state.tension = initialState.tension
-	state.friction = initialState.friction
-
-	output = {}
-	output.dx = state.v
-	output.dv = springAccelerationForState state
-
-	return output
-
-springIntegrateState = (state, speed) ->
-
-	a = springEvaluateState state
-	b = springEvaluateStateWithDerivative state, speed * 0.5, a
-	c = springEvaluateStateWithDerivative state, speed * 0.5, b
-	d = springEvaluateStateWithDerivative state, speed, c
-
-	dxdt = 1.0/6.0 * (a.dx + 2.0 * (b.dx + c.dx) + d.dx)
-	dvdt = 1.0/6.0 * (a.dv + 2.0 * (b.dv + c.dv) + d.dv)
-
-	state.x = state.x + dxdt * speed
-	state.v = state.v + dvdt * speed
-
-	return state
-
--- a/framer/Framer.coffee
+++ b/framer/Framer.coffee
@@ -28,6 +28,7 @@ Framer.LinearAnimator = (require "./Animators/LinearAnimator").LinearAnimator
 Framer.BezierCurveAnimator = (require "./Animators/BezierCurveAnimator").BezierCurveAnimator
 Framer.SpringDHOAnimator = (require "./Animators/SpringDHOAnimator").SpringDHOAnimator
 Framer.SpringRK4Animator = (require "./Animators/SpringRK4Animator").SpringRK4Animator
+Framer.LayerDraggable = (require "./LayerDraggable").LayerDraggable
 Framer.Importer = (require "./Importer").Importer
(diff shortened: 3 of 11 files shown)


koenbok:
I started working on the api proposal for this: https://github.com/koenbok/Framer/wiki/Proposed-Draggable-API

Other things we decided was to make Animator a mode of Simulator, as Animator is a deterministic Simulator.

PR #176, opened 2015-02-03
Allow base64-encoded image 'paths'

clintandrewhall:
This change allows base-64 encoded 'paths' to be placed in the JSON layer file, allowing single-page rendering of Framer prototypes.

### DIFF
--- a/framer/Importer.coffee
+++ b/framer/Importer.coffee
@@ -22,7 +22,7 @@ class exports.Importer
 
 		layersByName = {}
 		layerInfo = @_loadlayerInfo()
-		
+
 		# Pass one. Create all layers build the hierarchy
 		layerInfo.map (layerItemInfo) =>
 			@_createLayer layerItemInfo
@@ -43,7 +43,7 @@ class exports.Importer
 	_loadlayerInfo: ->
 
 		# Chrome is a pain in the ass and won't allow local file access
-		# therefore I add a .js file which adds the data to 
+		# therefore I add a .js file which adds the data to
 		# window.__imported__["<path>"]
 
 		importedKey = "#{@paths.documentName}/layers.json.js"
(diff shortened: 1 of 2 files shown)


framer/Importer.coffee
```diff
@@ -70,13 +70,16 @@ class exports.Importer
 		# Most layers will have an image, add that here
 		if info.image
 			layerInfo.frame = info.image.frame
-			layerInfo.image = Utils.pathJoin @path, info.image.path
-			
+			if info.image.path.indexOf('data:image') != -1
```
koenbok: I think it's a bit nicer to use https://lodash.com/docs#startsWith here.

koenbok: I think it's a bit more readable/correct to use https://lodash.com/docs#startsWith here.

framer/Layer.coffee
```diff
@@ -513,7 +513,7 @@ class exports.Layer extends BaseClass
 
 			# If the file is local, we want to avoid caching
 			# if Utils.isLocal() and not (_.startsWith(imageUrl, "http://") or _.startsWith(imageUrl, "https://"))
-			if Utils.isLocal() and not imageUrl.match(/^https?:\/\//) and @_cacheImage is false
```
koenbok: Here too.

koenbok:
So I guess the idea is to have one self contained json file with everything? I'm not completely sure what it's useful for, but I like experiments :-)

clintandrewhall:
Hi Koen!  Just a head's up: I sent you a message on Facebook, not sure if it went to your "other" inbox. :-)  Figured it would be faster just to explain over IM...

koenbok:
Can I still help with this?

jordandobson:
It would be great to get this in... I have a hacky way of doing this at the moment... but I'm not sure I need a JSON file either... does it really matter where they are stored?

koenbok:
Closing due to inactivity. I'm happy to help and land it at some point.

PR #194, opened 2015-03-26
Constructor options

edwinvanrijkom:
Dropping `exportable` in favor of native `enumerable` flag, and a `excludeFromProps` flag:

All properties defined using `@define` will now be kept in the properties dictionary; that is - unless they have their `excludeFromProps` flag raised. This dictionary is also used to forward constructor options to the respective properties, and on copying layers.

The `enumerable` flag (`true` by default) determines whether the property is enumerable from the JavaScript perspective; so whether or not it should show up on using `for (var field in obj)`.

### DIFF
--- a/framer/BaseClass.coffee
+++ b/framer/BaseClass.coffee
@@ -20,7 +20,7 @@ class exports.BaseClass extends EventEmitter
 
 		# See if we need to add this property to the internal properties class
 		if @ isnt BaseClass
-			descriptor.enumerable = descriptor.enumerable || !descriptor.hasOwnProperty('enumerable')
+			descriptor.enumerable ?= not descriptor.hasOwnProperty('enumerable')
 			descriptor.propertyName = propertyName
 
 			if not descriptor.excludeFromProps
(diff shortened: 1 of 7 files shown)


framer/BaseClass.coffee
```diff
@@ -19,44 +19,40 @@ class exports.BaseClass extends EventEmitter
 	@define = (propertyName, descriptor) ->
 
 		# See if we need to add this property to the internal properties class
-		if @ isnt BaseClass and descriptor.exportable == true
-			# descriptor.enumerable = true
+		if @ isnt BaseClass
```
koenbok: Style: in CoffeeScript, it's considered better to use `not`, `or` instead of `!` `||`

framer/BaseClass.coffee
```diff
@@ -80,14 +76,13 @@ class exports.BaseClass extends EventEmitter
 	keys: -> _.keys(@props)
 
 	@define "props",
+		excludeFromProps: true
 		get: ->
-			props = {}
-
+			value = {}
 			for k, v of @constructor[DefinedPropertiesKey]
```
koenbok: Style: couldn't we do _.pick(@, _.keys(@constructor[DefinedPropertiesKey]))

PR #199, opened 2015-04-06
Bugfix: Map over points array vs. non-existent point variable.

ryngonzalez:
In both `Utils.pointMax` and `Utils.pointMin`, `points` was being extracted from the args and then promptly not used in favor of a non-existent `point`variable:

```coffeescript
	points = Utils.arrayFromArguments arguments
	point = 
		x: _.max point.map (size) -> size.x
		y: _.max point.map (size) -> size.y
```


koenbok:
Thanks!

ryngonzalez:
No problem!

PR #202, opened 2015-04-08
Set correct index on scroll layer when using wrap

mattsjohnston:
When wrapping a layer using the ScrollComponent, the new scroll component layer should be inserted at the same index as the original layer being wrapped.


koenbok:
Great catch, thanks!

PR #203, opened 2015-04-12
Ability of adding and removing multiple subLayers

fliptheweb:


edwinvanrijkom:
Looks good! Thanks for contributing. What do you think @koenbok?

ajimix:
Shouldn't it be a new method? addSublayers and removeSublayers? In the future addSublayer and removeSublayer could be deprecated.

edwinvanrijkom:
I don't think it has to be a new method perse, for the proposed one is polymorph: it accepts both a single layer, as well as an Array of layers, so it isn't breaking the 'old' API.

ajimix:
Yes I know but it makes more sense as it's clearer but anyway, it's a matter of taste :)

edwinvanrijkom:
Yeah, true :-) The only rational argument that I can come up with is that it's clearest to name it after the most common use case, which probably is adding/removing a single Layer?

koenbok:
Hmm all valid, but such a small change I'm ok with whatever. But we will need some tests at the very least.

PR #221, opened 2015-05-29
Prevent page change in locked direction

mattsjohnston:
When using direction lock on page components, the component will successfully prevent you from dragging content in a locked direction, but if you are dragging your finger in the locked direction faster than the velocity threshold, when you release the page will still change to the next or previous page in the locked direction. This if statement makes sure that if that happens we just keep the user on the original page.


koenbok:
Am I correct in assuming this PR solves this too: https://www.facebook.com/groups/framerjs/permalink/687287008065040/

Thanks!

PR #228, opened 2015-07-03
Design @1x

peteschaffner:
Designing @1x is super important to me, and the more I use FS as a design tool, it becomes even more imperative (this means less ⌘I and more direct drawing in Framer). It is also extremely beneficial for cleaner communication between designers and engineers (app.coffee becomes a source of self-documenting design ... no more red-lines).

Here is [a test project](http://share.framerjs.com/qxva760srau4/) (with an augmented template Sketch file included) where you can see the changes in action.

Related: https://www.facebook.com/groups/framerjs/permalink/641286459331762/

### DIFF
--- a/extras/Studio.framer/index.html
+++ b/extras/Studio.framer/index.html
@@ -2,18 +2,11 @@
 <html>
 	<head>
 		<meta charset="utf-8">
-		
+
 		<meta name="format-detection" content="telephone=no">
 		<meta name="apple-mobile-web-app-capable" content="yes">
 		<meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
-
-		<script type="text/javascript">
-		// This automatically sets the right viewport scale on mobile devices
-		(function() {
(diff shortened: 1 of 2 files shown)


peteschaffner:
...I will write/update tests too if you are cool with this @koenbok ;)

peteschaffner:
I just realized that my changes to `index.html` will break things for existing projects that are designed in absolute pixels. I reverted those changes and added some logic to update the viewport meta tag when setting `DeviceComponent#contentScale`. This should make things compatible with existing projects and shouldn't require any of the Studio demos/templates to be reworked.

@koenbok what do you think about this?

koenbok:
Ooh this looks pretty interesting. I'll try and take a closer look tomorrow!

rssems:
Any news?

koenbok:
Hey @peteschaffner we're still pretty interested in solving this. Are you still thinking about finishing approach?

PR #230, opened 2015-07-15
Fix subLayers being affected by superLayer `html`

peteschaffner:
A sub-layer of a super-layer with `Layer#html` content set is not positioned correctly in the parent, as it is pushed down by the content. See [this example](http://share.framerjs.com/6ohd1s4lc4vl/).


koenbok:
Thanks man!

peteschaffner:
No prob! Thanks for merging :)

PR #232, opened 2015-07-19
Allow property setters to create sub-layers in construction

peteschaffner:
If you had a property that wanted to handle the logic of instantiating sub-layers during construction, it would throw an error (`TypeError: undefined is not an object (evaluating 'layer._subLayers.push')`) because the `_subLayers` property is set after calling super's constructor.

Though rare, this pattern [is necessary](https://github.com/peteschaffner/framer-material/blob/master/app.coffee#L377) when you have a subclass that requires sub-layer(s) and that also needs to be able to set properties on said sub-layer(s) when instantiated.

### DIFF
--- a/framer/Layer.coffee
+++ b/framer/Layer.coffee
@@ -47,8 +47,10 @@ class exports.Layer extends BaseClass
 
 	constructor: (options={}) ->
 
+		# Set needed private variables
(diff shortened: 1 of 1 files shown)


koenbok:
Fair.

PR #235, opened 2015-07-28
added siblingLayersByName() method

awt2542:
nice to keep things consistent with the subLayersByName() method.


koenbok:
Well why not. Thanks for adding tests too.

Issue #236, opened 2015-08-03
`Event.ImageLoaded` event never triggered in FramerJS?


hanfeisun:
Here is my code, which can be pasted into Framer Studio directly:

    layerA = new Layer()
    
    
    # Listen to the loading event
    layerA.on Events.ImageLoaded, -> 
        print "the image is loaded"
    
    layerA.on Events.ImageLoadError, 
        -> print "the image could not be loaded"
        
    layerA.image = "http://framerjs.com/static/images/home/app-icon.png"
    

However, I found the `the image is loaded` string is never printed.. Is it caused by cache? Or is it a bug.. How can I trigger the Events.ImageLoaded after an image is loaded?

jchavarri:
I think this is a duplicate of #190

jchavarri:
@koenbok I guess this one can be closed as well.

Issue #219, opened 2015-05-21
PageComponent resets position of layer


awt2542:
Seems like the PageComponent resets the position of a page. In the example below, I want each page to be centered inside the pageComponent (like the first page). Maybe I'm missing something.
```
page = new PageComponent 
    width: Screen.width
    height: Screen.height
    scrollVertical: false
   
for i in [0..2]
	temp = new Layer
	    width: page.width,
	    height: 450
	    superLayer: page.content
	    backgroundColor: Utils.randomColor()
	    midY: Screen.height/2
	if i > 0 then page.addPage(temp, "right")
```

PR #238, opened 2015-08-10
Fix issue #236 - `Event.ImageLoaded` event never triggered in FramerJS + Fix issue #219 "PageComponent resets position of layer"

jchavarri:
This PR targets issue #236: **`Event.ImageLoaded` event never triggered in FramerJS**

On the Layer class, the property `@events` was being checked in the `image` property definition in order to emit the `load` and `error` events once the image loading phase has ended. But this property `@events` is never created in the Layer class.

Renaming `@events` to `@_eventListeners` solves it.

It also solves issue #219 **PageComponent resets position of layer** by keeping the original position of the page before using `addPage` from a PageComponent

### DIFF
--- a/framer/Components/PageComponent.coffee
+++ b/framer/Components/PageComponent.coffee
@@ -98,7 +98,10 @@ class exports.PageComponent extends ScrollComponent
 			direction = "right"
 			throw new Error("#{direction} should be in #{directions}")
 
-		point = {x:0, y:0}
+		# For allowing pages added with 'addPage' to behave consistently with pages added 
+		# to the PageComponent using 'superLayer', keep the original page point
+		# so one of the two coordinates is left untouched after the page is added
+		point = page.point
 
 		if @content.subLayers.length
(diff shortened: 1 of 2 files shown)


jchavarri:
Sorry for the messy PR :( I'm pretty noob to Github contribs as you can see. I guess the best way to propose a fix is to create a branch 'fix-for-issue-236' and request from there, right?

If there's anything I can do to clean it up, let me know.

koenbok:
Hey this is great! Thanks so much!

tehfailsafe:
Did this get reverted? Pasting this in framer v47 never prints

```
layerA = new Layer()
layerA.on Events.ImageLoaded, -> 
    print "the image is loaded"

layerA.image = "http://framerjs.com/assets/static/images/frontpage/window-sketch.png"
```

jchavarri:
It looks like the same issue but the cause is different. It's being fixed on issue #292

Issue #86, opened 2014-06-22
Layer registration point vs originX/Y?


klebba:
The originX and originY properties are a bit confusing; I expected them to function as the registration point of the layer at first. Looking at how they are used with CSS transforms I now understand the difference, but I still want to affect the registration point of a layer for positioning purposes. This is what I'm talking about (Flash example) http://www.kirupa.com/developer/mx/rpoint.htm

Thanks!

awt2542:
Here's a better example.
1. Click on layerA - state is correctly switched.
2. Click on layerA again - state doesn't go back to default (i.e not looping). Try "x" instead, and it will work.
```
layerA = new Layer
layerA.states.add goRight: midX: 200
layerA.on Events.Click, -> @states.next()
```

PR #239, opened 2015-08-14
Fix issue #86: "Layer registration point vs originX/Y?"

jchavarri:
I've tracked issue #86 down. The problem is the properties of the "default" state are being ignored if the other states don't include "exportable" properties.

So from the issue example, "goRight" state has only one property, `midX`. But when going back to the "default" state, nothing happens. This is what is going on behind the scenes:

1. The "default" state only includes exportable properties (see [line 23 of LayerStates.coffee](https://github.com/koenbok/Framer/blob/master/framer/LayerStates.coffee#L23) and [line 95 of BaseClass.coffee](https://github.com/koenbok/Framer/blob/master/framer/BaseClass.coffee#L95))
2. Thus, state `default` does not include `midX` property because it's not exportable (see [line 244 of Layer.coffee](https://github.com/koenbok/Framer/blob/master/framer/Layer.coffee#L244))
3. When going back to `default` state, `midX` is not animated because of 1. and 2., and `x` is not included either because all the props from `default` are ignored (see [line 141 of LayerStates.coffee](https://github.com/koenbok/Framer/blob/master/framer/LayerStates.coffee#L141))

By removing line 141 of LayerStates, all the `default` properties are included in the animations, and thus the `default` state is restored completely. This doesn't affect other custom states animations, as only the properties from each state are iterated.

### DIFF
--- a/framer/LayerStates.coffee
+++ b/framer/LayerStates.coffee
@@ -138,7 +138,6 @@ class exports.LayerStates extends BaseClass
 		keys = []
 
 		for stateName, state of @_states
-			continue if stateName is "default"
 			keys = _.union(keys, _.keys(state))
 
 		keys
--- a/test/tests/LayerStatesTest.coffee
+++ b/test/tests/LayerStatesTest.coffee
@@ -121,4 +121,17 @@ describe "LayerStates", ->
 
 			layer.states.switch "stateA", {curve:"linear", time:0.1}
 
-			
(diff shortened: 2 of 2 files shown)


koenbok:
This looks great! Any chance you could add a small test?

jchavarri:
Done! :smiley:

koenbok:
Awesome, thanks so much!

PR #241, opened 2015-08-24
Adding tests for state next() function and reversing commit ecf3c54

jchavarri:
After trying to replicate the error pointed out on commit [ecf3c54](https://github.com/koenbok/Framer/commit/ecf3c54a895267f288d64b8217c60363fe2f24a5) and not being able to do so, I have added a test to cover the `next()` function of the `states` object. 

The test adds two states to a layer and iterates through them, changing some positioning and rotation values, and finally calls 'next()' one more time to go back to `default`. After every state transition it verifies that the values are correct.

The PR includes the reverse commit for [ecf3c54](https://github.com/koenbok/Framer/commit/ecf3c54a895267f288d64b8217c60363fe2f24a5) and the mentioned test to cover it.

### DIFF
--- a/framer/LayerStates.coffee
+++ b/framer/LayerStates.coffee
@@ -138,7 +138,6 @@ class exports.LayerStates extends BaseClass
 		keys = []
 
 		for stateName, state of @_states
-			continue if stateName is "default"
 			keys = _.union(keys, _.keys(state))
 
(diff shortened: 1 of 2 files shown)


koenbok:
Thanks so much @jchavarri !

PR #255, opened 2015-11-16
Added support for setting originZ on a Layer

nvh:
This adds support for setting originZ on a Layer. Normally originZ would defined as a [length](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin), because an element has no z-dimension, so it doesn't make sense to define it relatively. 

However, these pixels are defined the same way as perspective (which is the ['distance between the z=0 plane and the user'](https://developer.mozilla.org/en-US/docs/Web/CSS/perspective), so if a Layer has a superLayer that has a perspective, we can use that as a reference. This way originZ can be defined relatively, just as originX and originY.

In this pull request changing a superLayer recalculates the origin, but changing the perspective of the superLayer doesn't. As the perspective value isn't usually something you want to be varying (as opposed to [perspective-origin](https://developer.mozilla.org/en-US/docs/Web/CSS/perspective-origin)), this doesn't seem a big problem to me.

Example prototype using this branch is here: http://share.framerjs.com/k44wfwqgs73z/

### DIFF
--- a/framer/Layer.coffee
+++ b/framer/Layer.coffee
@@ -118,7 +118,7 @@ class exports.Layer extends BaseClass
 	# Matrix properties
 	@define "x", layerProperty(@, "x", "webkitTransform", 0, _.isNumber)
 	@define "y", layerProperty(@, "y", "webkitTransform", 0, _.isNumber)
-	@define "z", layerProperty(@, "z", "webkitTransform", 0, _.isNumber)
+	@define "z", layerProperty(@, "z", "webkitTransform", 0, _.isNumber, {relatedCssProperties: ['webkitTransformOrigin']})
 
 	@define "scaleX", layerProperty(@, "scaleX", "webkitTransform", 1, _.isNumber)
 	@define "scaleY", layerProperty(@, "scaleY", "webkitTransform", 1, _.isNumber)
@@ -135,7 +135,7 @@ class exports.Layer extends BaseClass
 
 	@define "originX", layerProperty(@, "originX", "webkitTransformOrigin", 0.5, _.isNumber)
 	@define "originY", layerProperty(@, "originY", "webkitTransformOrigin", 0.5, _.isNumber)
-	@define "originZ", layerProperty(@, "originZ", "webkitTransformOrigin", 0.5, _.isNumber)
+	@define "originZ", layerProperty(@, "originZ", "webkitTransformOrigin", 0, _.isNumber, {relatedCssProperties: ['webkitTransform']})
 
 	@define "perspective", layerProperty(@, "perspective", "webkitPerspective", 0, _.isNumber)
 
--- a/framer/Layer.coffee
+++ b/framer/Layer.coffee
@@ -133,7 +133,7 @@ class exports.Layer extends BaseClass
 
 	@define "originX", layerProperty(@, "originX", "webkitTransformOrigin", 0.5, _.isNumber)
 	@define "originY", layerProperty(@, "originY", "webkitTransformOrigin", 0.5, _.isNumber)
-	# @define "originZ", layerProperty(@, "originZ", "WebkitTransformOrigin", 0.5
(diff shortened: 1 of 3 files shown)


imp17:
NAVER - http://www.naver.com/
--------------------------------------------

cbox101@naver.com 님께 보내신 메일 <[Framer] Added support for setting originZ on a Layer (#255)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

jonastreub:
It would be cool to have have originZ for layers. When you change the existing originX and originY properties the layer won't change position until you transform the layer. I think people expect the same behaviour from originZ. Scale and skew of course won't change the due to z since layers don't have any depth. Its important for rotations though. It would be cool if the following code would make a cube:

```coffee
bg = new BackgroundLayer
	perspective: 600

cube = new Layer
	superLayer: bg
	backgroundColor: null
	midX: Screen.width / 2
	midY: Screen.height / 2
	clip: false
	style: webkitTransformStyle: "preserve-3d" # we should not have to do this

front = new Layer
	superLayer: cube

top = new Layer
	superLayer: cube
	originZ: -50
	rotationX: 90

right = new Layer
	superLayer: cube
	originZ: -50
	rotationY: 90

# etc.
```

To make that work we would probably have to use translateZ using the originZ amount before rotating and translating it back in the Z direction after rotating.

imp17:
NAVER - http://www.naver.com/
--------------------------------------------

cbox101@naver.com 님께 보내신 메일 <Re: [Framer] Added support for setting originZ on a Layer (#255)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

nvh:
I don't think I really understand. It's true that the originZ changes the scale of the layer, do you mean that the that scaling should be inverted so that the layer appears to be the same size? That sounds reasonable, I agree that without a transform the appearance of a layer should not change when you modify origin(X|Y|Z). On the other hand, if you change originZ and then rotate by 90 degrees, your layer is suddenly bigger or smaller than you would expect. That would be weird, right?

I created this implementation to be able to rotate along the X and Y axes while keeping a distance from the originZ, wouldn't you negate that by translating first as you suggest?

nvh:
What if originZ is relative to the layer's height? I guess that would provide the most expected results in most cases.

jordandobson:
@jonastreub - I brought that up when I was working on this: http://framerco.de/post/125466830249/3d-rotation-with-z-axis-depth-heres-an-example-of

It takes a bit of a hack on styles to get transforms on the Z-axis to work. It'd be great if the "preserve-3d" was enabled after you attempted to use some type of setting.

I have some other interesting explorations around this file if you'd like to see more. :D

nvh:
Here's an example using the current state of the PR: http://thewebrocks.com/demos/3D-css-tester/

jonastreub:
This looks great @nvh! I am currently experimenting adding a default perspective to each context and having all layers preserve 3D by default. That would make changes to certain properties like layer.z more clear but might also break some things.

jordandobson:
Jonas that would be awesome. This is something I found odd when I used it.
Also that I had to remove clipping.

jonastreub:
It seems `originZ` is rendered different by Chrome. We might be able to fix that by implementing originZ ourselves using a couple of transforms.

koenbok:
@jonastreub what do you want to do here?

jonastreub:
Not yet done but here is an alternative implementation using translateZ before and after rotating: https://github.com/koenbok/Framer/commit/6fd93913e370fe4e7c99cd9444c4ad0881134bc1

jonastreub:
I created a new [pull request](https://github.com/koenbok/Framer/pull/266) which should improve the overall 3D experience. Thanks @nvh for starting this!

nvh:
Cool!

koenbok:
This is now here: https://github.com/koenbok/Framer/pull/266

PR #261, opened 2015-12-01
Context freeze restore

koenbok:
For review.

This refactors a lot of the event handling code. We add our custom EventEmitter (with the exact same api) and base layer.events* and EventManager on it. Simplifies a lot of hairy code.

This also refactors context into a nice class. Some internal method names changed, but projects shouldn't break. Additionally you can now freeze/resume contexts.


framer/Context.coffee
```diff
@@ -1,122 +1,241 @@
...
+		fn()
 		Framer.CurrentContext = previousContext
 
+
+	##############################################################
+	# Freezing
+
+	freeze: ->
+
+		if @_frozenEvents
```
edwinvanrijkom: `if @_frozenEvents? ` is syntactically nicer here, 'cause it will return `true` when `@_frozenEvents` is `null` or `undefined`. Not that it matters functionally, in this case though :-)

framer/Context.coffee
```diff
@@ -1,122 +1,241 @@
...
+		@_element.classList.add("framerContext")
+
+		@__pendingElementAppend = =>
+			parentElement = @_parent?._element
+			parentElement ?= document.body
+			parentElement.appendChild(@_element)
+
+		Utils.domComplete(@__pendingElementAppend)
+
+
```
edwinvanrijkom: @koenbok Could this get some comments explaining what's going on? A context has its own `div`? Why does it need that when it's also attached to a Layer?

framer/EventEmitter.coffee
```diff
@@ -1,3 +1,94 @@
-EventEmitter = require "eventemitter3"
```
edwinvanrijkom: @koenbok It worries me a little that we're dropping a proven eventing lib and rolling with our own instead. Are you sure that we're not signing up for unforeseen bugs and performance issues?

framer/EventEmitter.coffee
```diff
@@ -1,3 +1,94 @@
...
+			@callListener(eventName, listener, arguments...)
+		@on(eventName, fn)
+		return @
+
+	listeners: ->
+		return _.clone(@[EventKey])
+
+	listenersForEvent: (eventName) ->
+		return [] unless @[EventKey]
+		return [] unless @[EventKey][eventName]
```
edwinvanrijkom: @koenbok These two can be merged, I think: `return [] unless @[EventKey]?[eventName]?`

framer/Context.coffee
koenbok: Fair.

framer/EventEmitter.coffee
koenbok: Oh nice one.

framer/EventEmitter.coffee
koenbok: Yeah, but event emitters are pretty easy, the code is well tested, and I've written them many times before.

Maybe this version a tiny bit slower because I don't do fancy case statements per n arguments, but we can optimize if it turns out to be a problem, and it really doesn't seem to be the case. Other than that, they are mostly the same.

I really think the tradeoff is worth it, having a better removal api, being able to remove the layer cruft, and having some sanity checking built in are good things.

framer/DOMEventManager.coffee
```diff
@@ -0,0 +1,41 @@
...
+			element._eventManagerId = EventManagerIdCounter++
+		
+		if not @_elements[element._eventManagerId]
+			@_elements[element._eventManagerId] = new DOMEventManagerElement(element)
+
+		@_elements[element._eventManagerId]
+	
+	reset: ->
+		for element, elementEventManager of @_elements
+			elementEventManager.removeAllListeners()
```
jchavarri: @koenbok I think this call to `removeAllListeners` won't work. The function in `EventEmitter` is now expecting an eventName to remove all its listeners. There was a similar function before in `EventManagerElement` but it was removed on the process.
I would propose to add a special case to `removeAllListeners`. If the param `eventName` is not set, it iterates over `@listeners` and remove them all.

koenbok: Good find. I'm going to do what you suggest :-)

framer/EventEmitter.coffee
koenbok: Update we're now back to EventEmitter3.

PR #263, opened 2015-12-05
Add build information and version number

koenbok:
This adds a few things:

- Automatic version information under Framer.Version, like build number, date, branch etc.
- Upgraded to @werckerbot Ewok (Docker) stack so it works for automatic builds.
- We ditched Bower for web based tests and instead get the packages directly from npm for simplicity.

More info about the Wercker stuff here: https://gist.github.com/koenbok/031bfe7831a8b270a1f7


package.json
```diff
@@ -19,21 +19,24 @@
...
+    "coverage-capture": "./test/coverage-capture",
     "gulp": "^3.8.11",
     "gulp-mocha-phantomjs": "^0.10.1",
+    "gulp-rename": "^1.2.2",
+    "gulp-template": "^3.1.0",
     "gulp-webpack": "^1.3.1",
-    "mocha-phantomjs": "^4.0.1",
-    "should": "^7.1.1",
-    "simulate": "^0.0.5",
-    "webpack": "^1.7.3",
```
edwinvanrijkom: Think this means that `make` will no longer work. Perhaps it'd make sense to remove the `Makefile`?

gulpfile.coffee
```diff
@@ -79,6 +87,32 @@ gulp.task "watcher", ->
...
+		task = gulp.src("framer/Version.coffee.template")
+			.pipe(template(info))
+			.pipe(rename({
+				basename: "Version",
+				extname: ".coffee"
+			}))
+			.pipe(gulp.dest("framer"))
+
+		callback(null, task)
+
```
edwinvanrijkom: Do I understand correctly that nowhere a version of `Versions.coffee` holding the actual set version values touches the sources? It only ends up in the build, correct? Hence we can take the existing hash, I suppose?

framer/Version.coffee.template
```diff
@@ -0,0 +1,7 @@
+# This gets automatically populated bij gulp
+
+exports.date = <%= date %>
```
edwinvanrijkom: It'll be impossible to rebuild an exact version replica when date is always build time's 'now'. So two builds functionally equivalent won't have the same file hash. Not sure if that matters though; might be handy to have the date for future use.

edwinvanrijkom:
@koenbok Yay! this is great. Left a few questions; and it looks like Wrecker is choking on the build.

framer/Version.coffee.template
koenbok: Hmm yeah good one, maybe for hashing. But I can't really think of something this will break right now.

gulpfile.coffee
koenbok: Yes because things get messed up when I insert anything after the webpack build, I just generate a file with version information before. I don't get the last sentence?

package.json
koenbok: Why not? It just re-ordered them, not removed.

koenbok:
Hmmm why doesn't this build?

koenbok:
Ok ugh, it almost works. But git doesn't work as expected on Wercker.

This reports version: `HEAD/525dae1-dirty build:112`
But it should be: `version-number/525dae1-dirty build:979`

koenbok:
@mies could you help us landing this next week?

PR #265, opened 2015-12-10
Fix local image path when using a custom device

jonastreub:
This fixes a bug where you cannot set device images for custom devices.


koenbok:
Do I need to review this @jonastreub ?

jonastreub:
Yes please!

framer/Components/DeviceComponent.coffee
```diff
@@ -73,6 +73,8 @@ class exports.DeviceComponent extends BaseClass
 		if @_setupDone
 			return
 
+		@_buildInDevices = _.keys Devices
```
koenbok: English: builtInDevices. Also we prefer `_.keys(Devices)`.

framer/Components/DeviceComponent.coffee
```diff
@@ -279,19 +281,24 @@ class exports.DeviceComponent extends BaseClass
...
 			return name
 
-		# If we're running Framer Studio and have local files, we'd like to use those
-		if Utils.isFramerStudio() and window.FramerStudioInfo
-			resourceUrl = window.FramerStudioInfo.deviceImagesUrl
-		
-		# If not, we want to get these image from our public resources server
-		else
-			resourceUrl = "//resources.framerjs.com/static/DeviceResources"
+		if @_deviceType in @_buildInDevices
```
koenbok: Why not early exit:

```
if @_deviceType in @_buildInDevices
  return name
```

test/tests/DeviceComponentTest.coffee
```diff
@@ -0,0 +1,34 @@
...
+			"screenHeight": 1000
+			"deviceImage": "images/framer-icon.png"
+			"deviceImageWidth": 800
+			"deviceImageHeight": 1203
+
+		# Set custom device
+		comp.deviceType = "test"
+
+		comp.deviceType.should.equal "test"
+		comp._deviceImageUrl(comp._deviceImageName()).should.eql "images/framer-icon.png"
```
koenbok: I'd like to see all permutations tested:

- In Framer Studio
- Not in Framer Studio
- Built in device
- Custom device

So, in total 2 x 2 tests.

framer/Components/DeviceComponent.coffee
koenbok: Also, in theory it's a waste to set this for every instance. You should make it a static global in the module.

framer/Components/DeviceComponent.coffee
```diff
@@ -278,20 +278,23 @@ class exports.DeviceComponent extends BaseClass
 		# If the image is externally hosted, we'd like to use that
 		if _.startsWith(name, "http://") or _.startsWith(name, "https://")
```
koenbok: Refactor this a bit so you end up with if...return if...return if...return.

test/tests/DeviceComponentTest.coffee
koenbok: Nice.

jonastreub:
Thanks for the feedback @koenbok! Code is much clearer now.

PR #266, opened 2015-12-12
Improving 3D, fixing translateZ issues

jonastreub:
# Overview

Work on improving 3D started with the [originZ proposal](https://github.com/koenbok/Framer/pull/255) by Niels van Hoorn.

#### new properties on `Layer`
- `originZ`, improves rotational transform capabilities
- `flat`, all layers now preserve 3D unless flat is set to true
- `perspectiveOriginX` and `perspectiveOriginY`, both default to 0.5 which is the center
- `backfaceVisible`, allows the back of the layer to be hidden

#### new properties on `Context`
- `perspective`, defaults to 1200 so all layer transforms in the z direction are observable without any setup
- `backgroundColor`, since backgroundLayers cover layers positioned on the negative z axis it is recommended to set a backgroundColor on the Context
- `perspectiveOriginX` and `perspectiveOriginY`, both default to .5 which is the center

The described properties of context are available on the `Screen` object for easy adjustments

```coffee
Screen.perspective = 800 # default is 1200

Screen.backgroundColor = "white" # default is transparent
```

### Example

Using the new `originZ` property creating a carousel in 3D is quite easy. No more hard calculations.

```coffee
for index in [0...18]
	l = new Layer
		originZ: -300
		rotationY: 20 * index
		backgroundColor: Color.random(.5)
	l.center()
```
screen shot 2015-12-12 at 12 15 39


koenbok:
Nice overview! Some questions I'd love to get answered before merging:

- What can you say about the performance impact of these?
- How does it impact Framer on desktop/mobile?
- What has been written about these properties in relation to performance?

jonastreub:
The performance is pretty good though the fps drops a bit earlier both on desktop and mobile. I didn't notice a difference in performance trying out existing prototypes like the [Weather example](http://share.framerjs.com/bxlo8z7vrarz/).

perf

test/tests/LayerTest.coffee
```diff
@@ -915,17 +939,17 @@ describe "Layer", ->
 			inputElement = _.first(inputElements)
 			inputElement.getAttribute("id").should.equal "hello"
 
-	describe "Force 2D", ->
+	# describe "Force 2D", ->
```
koenbok: Why is this uncommented?

framer/Context.coffee
```diff
@@ -50,6 +50,10 @@ class exports.Context extends BaseClass
 		@_parent = options.parent
 		@_name = options.name
 		
+		@perspective = 1200
```
koenbok: What happens if you chance these now. Nothing right? Are they supposed to be static?

framer/Context.coffee
```diff
@@ -62,6 +66,8 @@ class exports.Context extends BaseClass
 		@resetTimers()
 		@resetIntervals()
 
+		@perspective = @perspective
```
koenbok: ???

framer/Context.coffee
```diff
@@ -254,3 +260,44 @@ class exports.Context extends BaseClass
...
+			if _.isNumber(value)
+				@_perspective = value
+				@_element?.style["perspective"] = @_perspective
+				@_element?.style["webkitPerspective"] = @_perspective
+
+	_updatePerspective: ->
+		@_element?.style["webkitPerspectiveOrigin"] = "#{@perspectiveOriginX * 100}% #{@perspectiveOriginY * 100}%"
+
+	@define "perspectiveOriginX",
+		get: ->
```
koenbok: Maybe:

```
return @_perspectiveOriginX if _.isNumber(@_perspectiveOriginX)
return .5
```

framer/Context.coffee
```diff
@@ -254,3 +260,44 @@ class exports.Context extends BaseClass
...
+				@_element?.style["backgroundColor"] = new Color value.toString()
+
+	@define "perspective",
+		get: ->
+			return @_perspective
+		set: (value) ->
+			if _.isNumber(value)
+				@_perspective = value
+				@_element?.style["perspective"] = @_perspective
+				@_element?.style["webkitPerspective"] = @_perspective
```
koenbok: Double? We don't do that anywhere.

If we're going to do css prefixing this would be my least favorite way :-)

koenbok:
This is ready to merge in after our coming update.

koenbok:
This looks done. Let's go.

PR #269, opened 2015-12-14
emits 'slider.on value:change' when 'slider.value' is changed

marckrenn:
see http://share.framerjs.com/89mt081dckxo/


marckrenn:
I don't know what I'm doing :(

PR #244, opened 2015-09-09
Gesture support using Hammer.js

jchavarri:
**Ongoing work to be discussed, don't merge! :)**

Related to issue #107. Before I keep working on it, I would like to know your thoughts. 

Here's a first attempt to integrate Hammer with Framer, so we can recognize pinches, rotations, panning, etc. From the few tests I've done it seems to work well together with the existing events. I have tried to make both gesture and browser events transparent for the final user, even if behind the scenes they are two completely different things.

I see this as an exercise to explore how gesture events could be handled in Framer. I went with Hammer because they already deal with a lot of corner cases and gives a good idea on how to solve the gesture recognition in an elegant way. But maybe in the future this dependency could be removed by implementing the functionality in Framer?

This is an overview on how Hammer implements classes / concepts:

* `Input`: everything that can be consider a source of a gestural event. Subclasses: MouseInput, TouchInput, PointerEventInput, etc. This is also the class calculating velocities, directions and every extra info of the events.
* `Recognizer`: The gestures themselves. Subclasses: PanRecognizer, PinchRecognizer, etc. The base class contains a state machine to follow the gesture evolution.
* `Manager`: Assign handlers (or listeners) to recognizers based on a specific device input. It has similar responsibilities as Framer's `EventManagerElement`.

So what I've done for now is adding a GestureManager class that wraps Hammer functionality, and creates recognizers as they are requested depending on the gestures the layer is listening to.

Still to do: 

- Add the remaining events to the list (rotate, pan, etc)
- Implement `removeEventListener`: this one is a bit tricky: after removing a listener, we should remove the given gesture recognizer if there are no more events remaining of that type attached to the layer.
- Make complex examples: using ScrollComponent, PageComponent, sublayers, etc
- Write tests (are there any existing tests that target events, listeners and handlers?)

Side note: if this PR finally see the light, it would be very nice to integrate this [Touch Emulator](https://github.com/hammerjs/touchemulator) into Framer Studio, so users could emulate multi touch gestures using their keyboard. That would bring Framer a step closer to the iPhone Simulator!! :)

Finally, a small example that you can copy & paste:

```coffee
layerA = new Layer
	width: Screen.width
	height: Screen.height
	image: "http://lorempixel.com/900/900/"
# With draggable true some events are more difficult to trigger, but they work
layerA.draggable = false
initialScale = 0
layerA.on Events.PinchStart, (ev) ->
	# TODO layerA.origin = 
	initialScale = layerA.scale
[...]

PR #272, opened 2015-12-15
Simpler events

koenbok:
This allows for direct listener setting like `layer.onClick` and `layer.onChange "x", -> ...`. It also cleans some dom event listeners.


jchavarri:
I like this! Is this a definitive change or a test? Asking for [a friend](https://github.com/koenbok/Framer/pull/244) :)

koenbok:
Ha, good catch! Yes this is definitive.

PR #276, opened 2015-12-15
Sketch update preparations

koenbok:
So we can import at 1x/2x/3x.


koenbok:
Wercker is on it's ass. But the tests run local.

Issue #157, opened 2015-01-18
Layout constraint system


koenbok:

jordimo:
this would be awesome. going to give it a try

PR #281, opened 2015-12-31
Experiment: Flexbox for Framer

jchavarri:
Related to issue #157, I have been researching lately about constraints systems: iOS autolayout implementation, cassowary, Slalom etc. and how they could be integrated into Framer. IMHO these frameworks go somehow against Framer philosophy: layout constraint systems introduce great challenges in both language syntax (visual format, verbosity) and rules conflict handling.

I then started looking for other solutions to easily add a basic set of "fluid rules" to layers in Framer. Through [this article (Layout Abstractions)](http://khanlou.com/2015/02/layout-abstractions/) I found this awesome [css-layout](https://github.com/facebook/css-layout) library, that has been extracted from React Native. I instantly liked it for several reasons:

- It's based the flexbox and box model, but implements a subset of it. Many web developers are used to those models already, and they are vastly documented
- It's backed by Facebook
- It can be easily added to Framer syntax, just by adding some properties to the `Layer` class

I have created a basic example that you can use with this PR. Please let me know your ideas, and if you think this could be a useful addition to Framer.

**How it looks**
flexbox-framer

**Example code**
 ```coffeescript
bg = new BackgroundLayer
	backgroundColor: "#eee"

createMainContainer = ->
	
	mainContainer = new Layer
		backgroundColor: "#FFF"
		name: "mainContainer"

	mainContainer.states.add
		moveX:
			x: 50
	mainContainer.states.animationOptions =
	    curve: "spring(100, 10, 0)"

	mainContainer.on Events.Click, ->
		mainContainer.states.next()

	mainContainer.layout
		flexDirection: 'column'
		flex: 1

	createTopLayer(mainContainer)
	createMiddleLayer(mainContainer)
	createBottomLayer(mainContainer)

createTopLayer = (superLayer) ->

	topLayer = new Layer
		superLayer: superLayer
		backgroundColor: "#877DD7"
		name: "topLayer"
		html: "Top layer<br/>flex: 0.2 (20% of the screen height)<br>minHeight: 40)"
	topLayer.style.textAlign = "center"
	topLayer.style.fontSize = "18px"
	
	topLayer.layout
		flex: 0.2
		minHeight: 40

createMiddleLayer = (superLayer) ->

	middleContainer = new Layer
		superLayer: superLayer
		backgroundColor: "#eee"
		name: "middleContainer"
		html: "Middle container. flex: 0.8, flexDirection: row, paddingTop: 50"

	middleContainer.style.color = "gray"
	middleContainer.style.fontSize = "18px"
		
	middleContainer.layout
		flex: 0.8
		flexDirection: "row"
		paddingTop: 50

	leftPanel = new Layer
		superLayer: middleContainer
		backgroundColor: "#28AFFA"
		name: "leftPanel"
		html: "flex: 0.2"
	leftPanel.style.fontSize = "18px"

	leftPanel.layout
		flex: 0.2

	centerPanel = new Layer
		superLayer: middleContainer
		backgroundColor: "white"
		name: "centerPanel"
		html: "This layer has flex: 0.8, margin: 20, minWidth: 400. <br/>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut quis aliquet neque. Nunc cursus ligula nec commodo laoreet. Aliquam congue leo in lobortis volutpat. Sed hendrerit, tortor eu lacinia congue, eros sapien venenatis felis, at finibus neque purus at ipsum. Aliquam a hendrerit augue, eu vulputate velit. "

	centerPanel.style.color = "#999"
	centerPanel.style.padding = "32px"
	centerPanel.style.fontSize = "24px"
	centerPanel.style.fontWeight = "400"
	centerPanel.shadowY = 1
	centerPanel.shadowBlur = 2
	centerPanel.style.boxShadow = "0 1px 3px rgba(0,0,0,0.2)"
	  
	centerPanel.layout
		flex: 0.8
		margin: 20
		minWidth: 400

	rightPanel = new Layer
		superLayer: middleContainer
		backgroundColor: "#28AFFA"
		name: "rightPanel"
		html: "flex: 0.2"
	rightPanel.style.fontSize = "18px"

	rightPanel.layout
		flex: 0.2

createBottomLayer = (superLayer) ->

	bottomLayer = new Layer
		name: "bottomLayer"
		backgroundColor: "#7DDD11"
		superLayer: superLayer
		html: "Bottom layer. height: 80, bottom: 0"

	bottomLayer.layout
		bottom: 0
		height: 80

createMainContainer()
```


jchavarri:
Some TODO tasks:
- Fields validation, not only before adding to tree, but also before assigning to ‘rules’
- Optimise the tree creation and update
- Performance tests

And some open questions:
- Should properties (`flex`, `margin` etc) belong to `Layer`?
- Animations??
- How do `relative` and `absolute` from css-layout affect Framer layers?

jchavarri:
Another update! :)

So I have keep experimenting with this flexbox thing, and I find it amazing how quickly one can understand and use the flex model.

The most recent changes are:
- Redefined completely how the layout tree was stored and treated on `LayerLayout`. Now each layer keeps track of its layout, and notify the parent layer when an update is in order
- Flex properties are now part of `Layer`. This can be a big deal, but that was the quickest way I could think of to test how layout animations performed
- **Animations!** by adding the layout properties to `Layer`, I got all those Framer curves working for free! (more below)
- There's a `layout` property on `Layer`, instance of class `LayerLayout` that is in charge of everything layout related. 
- This `layout` property has itself a `curve` property that allows to configure a specific animation curve for new layouts. So *each layer* can decide how it will animate its layout updates, or to not animate them at all if `curve` is undefined.
- This "flex mode" keeps being 100% compatible with the absolute positioning one

**Performance**
There are actually two animation "modes" that we can evaluate:

1. Animate `layout` properties (`flex`, `fixedWidth`, etc): this is quite CPU intensive. I made a test where I was adding 100 layers simultaneously and it definitely hurts on the iPhone. Sometimes even made Mobile Safari Crash. However, is this a common case? I saw that if there's only 1 to 5 layers animating at a precise moment, even if they are affecting other layers, the phone seems to handle it. Maybe @koenbok can point me how to test performance better?
2. Animate using the `curve` property of `LayerLayout`: this kind of animations are just a frame animation and thus can be perfectly handled (for example, animate from `row` to `column` layout).

**Syntax**
As I said at the beginning, I find flexbox declarative approach very simple to use. Maybe it doesn't allow for as complex motion description as iOS AutoLayouts or Slalom, but I think that anyone can quickly understand and start playing with it.

[Here](https://github.com/jchavarri/framer-flex)'s the code for the example uploaded above, the most interesting part is in the `Content` class, where the `curve` property is set. You can play with it [here](http://jchavarri.github.io/framer-flex/).

jchavarri:
After reviewing how [react-canvas](https://github.com/Flipboard/react-canvas/blob/master/lib/layoutNode.js) integrates css-layout, and only updates the node tree when a new frame is needed, I realized that I was calculating the layout tree every time a property was updated. This, of course, had a great impact on performance. This is the `Perf.framer` results for every case (on a Macbook Pro):

a) `master` Framer branch, without css-layout, animating `x` and `y` props:

screen shot 2016-01-23 at 22 24 35

b) Using css-layout without optimization, animating `left` and `top` props (computes layout tree for every property update)

*(yes, that's 40 layers at 29.3fps)*

screen shot 2016-01-23 at 22 26 04


c) Using css-layout, but only computes the tree everytime the Framer.Loop `update` event is called, animating `left` and `top` props:

screen shot 2016-01-23 at 22 24 57

So I think this solves the performance problems mentioned in the previous comment. On the iPhone it works fine now as well, rendering up to 120 layers at 30fps (same results as the non-flex version), and Mobile Safari doesn't crash anymore.

I will update a commit soon with these changes. Of course this still needs a lot of work, but now that performance is better, it's more interesting to explore a bit further.

nvh:
This is a really interesting experiment, thanks for the hard work!
However, I don't think we currently want css-layout as a dependency of the Framer Library. Could you maybe wrap this in a module? That would be super cool.

PR #291, opened 2016-01-14
Gestures refactor to layer.gestures

koenbok:
Gestures and events are really two different concepts in Framer:

- **Events** are things that actually happen and are visually modifying things.
- **Gestures** are a more low level to watch for input and respond to them yourself. 

Or, gestures are just a low level form of events that you can build on. Think GestureRecognizer on iOS.

Respond to a draggable/pinchable/rotatable layer:

```coffee

layer = new Layer
layer.draggable.enabled = true
layer.draggable.pinchable = true
layer.draggable.rotatable = true

layer.onPinch (event) -> ...
layer.onRotate (event) -> ...
```

Respond to a pinch gesture (so you can do something with it yourself):

```coffee

layer = new Layer
layer.gestures.onPinch (event) -> ...
```


koenbok:
cc @jchavarri

framer/DOMEventManager.coffee
```diff
@@ -9,21 +9,14 @@ EventManagerIdCounter = 0
...
 	constructor: (@element) ->
-		@_elementGestureManager = new GestureManagerElement(@element)
 
 	addListener: (eventName, listener, capture=false) ->	
 		super(eventName, listener)
-		if Events.isGestureEvent eventName
-			@_elementGestureManager.addEventListener(eventName, listener, capture)
-		else
-			@element.addEventListener(eventName, listener)
+		@element.addEventListener(eventName, listener, false)
```
jchavarri: I know that this is not PR related, but why does the event need to be added to the DOM and also to EventEmitter? Is it to have the `listeners` and all the other handful EventEmitter functions?

framer/GestureManager.coffee
```diff
@@ -1,12 +1,55 @@
+{_} = require "./Underscore"
 Hammer = require "hammerjs"
 
-class exports.GestureManagerElement
+Utils = require "./Utils"
+{EventEmitter} = require "./EventEmitter"
+{Gestures} = require "./Gestures"
 
-	constructor: (@element) ->
+class exports.GestureManager extends EventEmitter
```
jchavarri: :+1:

framer/DOMEventManager.coffee
koenbok: Exactly. The dom removeEventListener is very picky and needs an exact reference to the handler. And if I account for them here I get nice things like `removeAllEventListeners()` for free.

framer/Gestures.coffee
```diff
@@ -0,0 +1,47 @@
...
+Gestures.Pinch = "pinch" # This event includes all the other Pinch events
+Gestures.PinchStart = "pinchstart"
+Gestures.PinchMove = "pinchmove"
+Gestures.PinchEnd = "pinchend"
+Gestures.PinchCancel = "pinchcancel"
+Gestures.PinchIn = "pinchin"
+Gestures.PinchOut = "pinchout"
+
+# Press
+Gestures.Press = "press"
```
jchavarri: I just took Hammer naming, but I was thinking that maybe "Press" is not very helpful as the gesture name. Maybe "LongPress" or "PressHold" is more meaningful?

jchavarri:
It's much cleaner now! I see very clearly that DOMEventManager was not the place to handle gestures.

Maybe now that `gestures` is a property of `Layer`, using `layer.gestures.on Gesture.Pan` is a bit more cumbersome than the previous `layer.on Events.Pan` but I think that can be solved using the new `onSwipe` & the other methods.

koenbok:
Ok all is implemented and ready for testing.

@jchavarri we have the same `layer.gestures.onPinch` format too.

jornvandijk:
Boom!

jchavarri:
@koenbok I have a couple of commits to propose, but no permissions as i'm not the owner of this PR. 

Check [here](https://github.com/koenbok/Framer/compare/gestures...jchavarri:gestures)

The only important addition is a check if a specific recognizer already exists for a given layer, so they don't get added multiple times when setting up multiple listeners with the same gesture.

jchavarri:
@jornvandijk we're close!!  😁

koenbok:
Thanks! I merged those in.

Issue #290, opened 2016-01-14
ImageLoaded events are not triggered


jchavarri:
`ImageLoaded` event listener is never called.
FB post: https://www.facebook.com/groups/framerjs/permalink/727413844052356
Example: http://share.framerjs.com/elaodngxso5n/

jchavarri:
PR #292 has been merged. Closing.

PR #292, opened 2016-01-14
Events.ImageLoaded not being triggered

jchavarri:
I'm including a failing test for issue #290 

From what I've seen, by using getEventListeners on Chrome console, the event is added to the DOM node, but the listener is never called. 

@koenbok Any ideas where I could look?


jchavarri:
@koenbok This is ready for review. Just a note: I have added a test to cover Events.ImageLoadError. The "problem" is that now an 404 error is being printed to the console whenever the test suite is run. Maybe there's another way to cover for that case without the error being printed?

koenbok:
Perfect! Thanks a ton!

jchavarri:
wow, that was quick! 😮

PR #296, opened 2016-01-20
Touch Emulator

koenbok:
This adds a touch emulator on desktop. It is enabled by default on non-touch devices.

Use `alt` to pinch and `alt` + `cmd` to pan.

This also changes the gesture api to:

- `layer.pinchable.enabled = true` like draggable
- `layer.on Gestures.Pinch` for raw gestures


framer/Extras/TouchEmulator.coffee
```diff
@@ -0,0 +1,206 @@
...
+			
+	touchEvent.touches = touchEvent.changedTouches = touchEvent.targetTouches = touches
+
+	target.dispatchEvent(touchEvent)
+
+cancelEvent = (event) ->
+	event.preventDefault()
+	event.stopPropagation()
+
+class TouchEmulator extends BaseClass
```
jchavarri: This!

framer/Gestures.coffee
```diff
@@ -1,5 +1,6 @@
 # Gesture events
 Gestures = {}
+Gestures._prefix = "gesture:"
```
jchavarri: Maybe `.prefix` without underscore as it's public?

framer/GestureManager.coffee
```diff
@@ -207,11 +177,13 @@ splitStr = (str) ->
 
 addEventListeners = (target, types, handler) ->
 	splitStr(types).map (type) ->
+		#console.log("hammer.addEventListener", type)
 		Framer.CurrentContext.domEventManager.wrap(target)
```
jchavarri: One day, I hope I'll understand this :(

jchavarri:
@koenbok If I can help somehow, let me know! :)

koenbok:
Haha so... much... work... and we changed the api again because we didn't like it.

It's in the latest beta now, you can try an example here: http://share.framerjs.com/rjk8t2bgmsra/

framer/Extras/TouchEmulator.coffee
koenbok: :-)

framer/Gestures.coffee
koenbok: Well it's for internal use only. It's how I distinguish between Events.* and Gestures.*.

framer/GestureManager.coffee
koenbok: Yeah I know. This is some nasty hacking. The end result is simple though: a list of all dom event listeners so we can manage them, because there is no way in js to ask which listeners are connected to a node, or any nodes.

jchavarri: Mm.. I was using `getEventListeners` for this, but I just discovered it's only available in Chrome. And it only expects 1 node, not all of them. http://stackoverflow.com/a/16544813/617787

jchavarri:
I see!! But it's looking really great. I like the `pinchable` property, it can't get any easier, I guess. :)

Re: the beta link, some things I noticed:
1. If you Alt+click or Alt+Cmd+click just after opening, it gives the following error:
```
TouchEmulator.coffee:76 Uncaught TypeError: Cannot read property 'x' of undefined
TouchEmulator.showTouchCursor	@	TouchEmulator.coffee:76
TouchEmulator.keydown	@	TouchEmulator.coffee:104
(anonymous function)	@	TouchEmulator.coffee:1
```
In this case, the line that is crashing is: 
```
	showTouchCursor: ->
		@touchPointLayer.animateStop()
		@touchPointLayer.midX = @point.x
```
Sometimes it shows this other one:
```
Uncaught TypeError: Cannot read property 'x' of undefined
Utils.pointSubtract @ Utils.coffee:563
TouchEmulator.panPoint @ TouchEmulator.coffee:200
TouchEmulator.mousemove @ TouchEmulator.coffee:166
(anonymous function) @ TouchEmulator.coffee:1
```
That corresponds to:
```
Utils.pointSubtract = (pointA, pointB) ->
	return point =
		x: pointA.x - pointB.x
		y: pointA.y - pointB.y
```
2. For the two pointers that appear, I'd suggest to show them separated just after the user hits Alt, so she sees very clearly that the pinch action has started
3. Sometimes it seems that when pinching, it doesn't keep the original scale value so the layer gets larger after every pinch gesture

I could help tomorrow night with any of these. :+1:

PR #297, opened 2016-01-24
Rewrite gesture support

koenbok:
So this is never fun, but ultimately Hammer got us more trouble than it solved.

- Hard to manage events attached to dom without hacks.
- Hard to understand gesture dependencies.
- Did not play nice with other events, leading to hard to solve bugs (for example with draggable).

I rewrote a simple version of hammer with most events we need.


jchavarri:
Sorry to hear that! :( I remember thinking some time ago about how Hammer would introduce a huge dependency to Framer, but i guess I got excited when I saw all the examples that I was testing were working. So the good news are that this dependency is not there any more.

jchavarri:
Can I help wrap this up? Testing, debugging...

koenbok:
Yep, same here. I initially thought it was better to build on it too. But sometimes the accumulation of issues leads you here :-)

I'd love you to do a sanity check on this PR when I finish it and test it a bit.

jchavarri:
Sure! ping me when it's ready

koenbok:
We'll start the merge so we can ship betas

koenbok:
PING @jchavarri

framer/GestureInputRecognizer.coffee
```diff
@@ -0,0 +1,448 @@
...
+GestureInputForceTapMobile = 0.7
+GestureInputForceTapMobilePollTime = 1/30
+
+{DOMEventManager} = require "./DOMEventManager"
+
+class exports.GestureInputRecognizer
+	
+	constructor: ->
+		@em = new DOMEventManager()
+		@em.wrap(window).addEventListener("touchstart", @touchstart)
```
jchavarri: When I saw this, I thought it was a bit extreme. But after giving some thought, I don't see any other way to do it. I checked Hammer code and they use this technique too (although they add some extra elements besides `window`)

framer/GestureInputRecognizer.coffee
```diff
@@ -0,0 +1,448 @@
...
+	swipe: (event) =>
+		@_dispatchEvent("swipe", event)
+		@swipedirection(event)
+
+	swipeend: (event) =>
+		@_dispatchEvent("swipeend", event)
+
+	# Direction swipe
+
+	swipedirectionstart: (event) =>
```
jchavarri: This is more API related, but it is weird for me (as a Framer user) to have `SwipeStart` and `SwipeEnd`. I mean, a swipe event imo should only trigger once (at the end of the gesture). Having Swipe events triggered while the gesture is ongoing make them identical to Pan, right?

framer/GestureInputRecognizer.coffee
```diff
@@ -0,0 +1,448 @@
...
+		return unless event.offsetDirection
+		return if @session.started.swipedirection 
+		@session.started.swipedirection = event
+		direction = @session.started.swipedirection.offsetDirection
+		@_dispatchEvent("swipe#{direction}start", event)
+
+		swipeEdge = @_edgeForSwipeDirection(direction)
```
jchavarri: If this is what I think it is, that's **awesome** :+1: :+1: 

Btw I am not able to trigger any 'edge' events using the Beta on the laptop, but maybe I'm doing it wrong.

framer/GestureInputRecognizer.coffee
```diff
@@ -0,0 +1,448 @@
...
+		return touchEvent
+
+	_dispatchEvent: (type, event, target) ->
+		touchEvent = @_createEvent(type, event)
+		target ?= event.target
+		target.dispatchEvent(touchEvent)
+
+	_getVelocity: (events) ->
+
+		return {x:0, y:0} if events.length < 2
```
jchavarri: I tried with `Pinch` and `Pan` but velocity is always 0. Is that normal?

jchavarri:
Man, this release is going to be huge! it's very clear when you see the code that it's been an intense sprint. Reimplementing Hammer in less than 2 days, :white_check_mark: !  :clap: :clap: 

Things I like as a user:
- Edge events (I could use it for that "framered" UINavigationController I had around)
- Force touch
- The gesture events have an API under "Framer control", which means it can be documented and used more by users
- `layer.pinchable`! 

One last thing, on the Beta, the main cursor and the secondary one for pinching have a different size. I guess you are aware, but just in case...

image1-

framer/GestureInputRecognizer.coffee
koenbok: Not really. Our use case was to build the notifications drawer on iOS. Although that's an edge swipe it's more or less the same. But you really need all of those events to implement it.

framer/GestureInputRecognizer.coffee
koenbok: Fixed on master.

framer/GestureInputRecognizer.coffee
koenbok: Ok they should work (swipe from device edge) but I'll check again.

koenbok:
Hey that is odd I'll look into this!

PR #310, opened 2016-02-10
Fix for mouseWheelEnabled at initialization

jordandobson:
Regardless if you pass the option when you create the component it will always default to false. This should pass the option in when it's initialized so it works as expected.


jordandobson:
Can we pull this one in guys? It's a tiny fix. :dancers:

koenbok:
lgtm

PR #317, opened 2016-02-22
Allow Layers to have both an image and a background color.

eelco:


jornvandijk:
👊

koenbok:
We should add a simple test for this behavior :-)

eelco:
Good point, I’ll look into this tomorrow!

eelco:
Is this a good one for the Future branch? Happy to rebase.

PR #320, opened 2016-03-01
Allow borderRadius to be set to an Array

eelco:
This allows the `borderRadius` of a `Layer` to be set to an array, to have different radiuses for different corners.

TODO:
- [ ] Add tests

### DIFF
--- a/framer/Layer.coffee
+++ b/framer/Layer.coffee
@@ -274,8 +274,11 @@ class exports.Layer extends BaseClass
 
 		set: (value) ->
 
-			if value and not _.isNumber(value)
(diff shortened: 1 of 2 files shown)


jordandobson:
You can already do this by passing a string. I like where you are going
here though.

eelco:
Good point, I noticed that it currently has a ‘compatibility fix’ in the code, I assume that’s for passing strings? Would it make sense to tighten the API up and remove this?

eelco:
Post-future.

jordandobson:
👍🏼

PR #324, opened 2016-03-04
Design @1x

peteschaffner:
I'm back :)

@koenbok I simplified my approach from #228, but now need a little help from you to get this in an acceptable state to merge.
## Must-haves
- [ ] when importing from Sketch/Photoshop, honor `Framer.Device.contentScale` by drawing all imported layers at the appropriate size (divided by the scale factor)

I'm a bit confused with the existing behavior because if I import a design drawn at 1x using the new importer scaling feature, all my layer dimensions are scaled by that factor, yet the `layerFrame` and `image.frame` values in the generated JSON are relative (at 1x) values.
## Nice-to-haves
- [ ] trigger a refresh of the device context when changing orientation so that it is not necessary to reload the preview (this would be great to trigger on device as well if it is possible)
- [ ] make the viewport meta tag generator DRY (maybe create a build step for including a factored-out version of the function both in `index.html` and `DeviceComponent.coffee`)
- [ ] consider making a less explicit API that requires direct knowledge of device scale factors and make something more implicit like `DeviceComponent#relativeUnits` and then do the extra legwork to add the appropriate `contentScale` properties to the built-in device configurations
- [ ] test(s) for fullscreen and simulated device contexts

### DIFF
--- a/framer/Components/DeviceComponent.coffee
+++ b/framer/Components/DeviceComponent.coffee
@@ -388,10 +388,7 @@ class exports.DeviceComponent extends BaseClass
 		@_contentScale = contentScale
 
 		if animate
-			@content.animate _.extend @animationOptions,
-				properties: {scale: @_contentScale}
-		else
-			@content.scale = @_contentScale
+			@content.animate @animationOptions
 
 		@_update()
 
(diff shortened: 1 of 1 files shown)


nvh:
@peteschaffner I'm doing some cleanup, is this still relevant? If yes, could you explain the situation and provide some examples?

nvh:
As we added support for pixel ratio independent design in Framer Design, I'm closing this. Feel free to open up a new issue or pull request to discuss it further.

PR #336, opened 2016-04-07
added layer.find() and Utils.layerMatchesSelector()

awt2542:
This adds the ability to find and return a layer based on its name property. The find() method accepts CSS-like selectors which are parsed by the Utils.layerMatchesSelector() method.

Examples:
layer.find('button') - return the descendant named button.
layer.find('card1 > button') - return the descendant named button, that's a direct child of card1.
layer.find('container button') - return the descendant named button that's a descendant of container.
layer.find('button*') - return the descendant whose name starts with 'button', eg. button1.

| Selector | Result |
| --- | --- |
| A | Any layer named A |
| A B | Any layer named B that is a descendant of a layer named A (that is: a child, or a child of a child, etc.) |
| A > B | Any layer named B that is a child (i.e. direct child) of a layer named A |
| A, B | Any layer named A or any layer named B |
| * | Any layer (wildcard character) |

The selector parser could be used for other functions in the future. Something like Utils.find(selector) would be a nice complement to layer.find() for when you want to search all layers in the project, and not just descendants of a specific layer.


jordandobson:
👏🏻

koenbok:
Hey super cool, I totally missed this!

koenbok:
I think it should be a class method right? Like `Layer.find(...)`.

nvh:
I think we should merge this, but with a couple of changes:
- [ ] Rename `find` to `select`
- [ ] Add a `selectLayer` to `Context`
- [ ] Make `Layer.select` work on `Framer.CurrentContext`
Will you be able to pick this up @awt2542?

awt2542:
@nvh Not sure how to make it work on CurrentContext, but I think it fixed the rest. Can you have a look?

nvh:
Closing in favor of #456

PR #342, opened 2016-04-13
Only converting properties whose name end in 'color'

nvh:
This caused a weird bug where the value of `html` would be replaced by a color string when in contained a color string (like "add")

### DIFF
--- a/framer/LayerStates.coffee
+++ b/framer/LayerStates.coffee
@@ -177,6 +177,8 @@ class exports.LayerStates extends BaseClass
 
 		# TODO: Maybe we want to support advanced data structures like objects in the future too.
 		for k, v of properties
+			# We check if the property name ends with color, because we don't want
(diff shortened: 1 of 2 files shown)


framer/LayerStates.coffee
```diff
@@ -179,7 +178,7 @@ class exports.LayerStates extends BaseClass
 		# TODO: Maybe we want to support advanced data structures like objects in the future too.
 		for k, v of properties
 
-			if _.isString(v) && Color.isColorString(v)
+			if _.isString(v) && _.endsWith(k.toLowerCase(),'color') && Color.isColorString(v)
```
koenbok: - We use `"` in the Framer source. 
- We favor `and` over && in cs.

Ideally I would add some comments here why we check the key name.

koenbok:
👌

PR #346, opened 2016-04-20
Add support for ‘native’ gestures.

eelco:
This adds support for ‘native’ gestures, e.g. those made on a trackpad. The `GestureEvent` used abstracts over touch points, so that information is missing, but this does allow for pinching in the (desktop) browser.

Needs more testing and maybe a different approach (e.g., don’t use this on mobile?)


koenbok:
Is this waiting for @koenbok ?

eelco:
Hmm, not specifically, but it’s waiting for more thoughts and considerations.

I think that for Desktop it’s nice to have native pinch, scroll, pan gestures too, but I’m not sure yet what the relation with the other gestures / touch emulator should be. My hunch would be: only do this on Desktop and only if the touch emulator is disabled. That might be too narrow a use case to though. Thoughts welcome :)

PR #353, opened 2016-04-30
Hints

koenbok:
Simple hints that show you what interactive areas are. This helps when you are looking at a prototype, but you have no idea where to click.

#### API

Every layer has a `showHint` function that you can override for custom looks. For example a ScrollComponent with only vertical scrolling might have a special hint. The methods cascades down to children if you call super.

- `Framer.Extras.Hints.enable()`
- `Layer.shouldShowHint()` returns bool if the layer has interactive things.
- `Layer.showHint()` shows the actual hint for this layer, customizable.


framer/Layer.coffee
```diff
@@ -121,10 +121,16 @@ class exports.Layer extends BaseClass
...
+		# Set some calculated properties
 		# Make sure we set the right index
 		if options.hasOwnProperty("index")
 			@index = options.index
 
+		# x and y always win from point, frame or size
+		for p in ["x", "y", "width", "height"]
+			if options.hasOwnProperty(p)
+				@[p] = options[p]
+
```
koenbok: Hmmm this should not be here anymore I think.

framer/Utils.coffee
```diff
@@ -892,6 +892,32 @@ Utils.pointInPolygon = (point, vs) ->
 		j = i++
 	inside
 
+Utils.frameIntersection = (rectA, rect) ->
```
koenbok: Todo: cleanup

PR #363, opened 2016-05-29
Make all gesture event points screen based

koenbok:
Announcement: https://www.facebook.com/groups/framerjs/permalink/891074961019576/


jordandobson:
😎

PR #372, opened 2016-06-23
Feature/image load cancel

nvh:
The problem: when setting using the image loader and setting the image to null afterwards, the loading wasn't cancelled, so the image would be set to the initial image when the loading finished.

This PR fixes that, and adds the possibility to listen to image cancel events.

I've changed the check for listeners from domEvent listeners to layer listeners, but maybe this has undesired side-effects that aren't currently clear to me.

### DIFF
--- a/framer/Events.coffee
+++ b/framer/Events.coffee
@@ -42,6 +42,7 @@ Events.Scroll = "scroll"
 # Image events
 Events.ImageLoaded = "image:load"
 Events.ImageLoadError = "image:error"
+Events.ImageLoadCancelled = "image:cancelled"
 
 # Add all gesture events
 _.extend(Events, Gestures)
--- a/framer/Events.coffee
+++ b/framer/Events.coffee
@@ -40,9 +40,9 @@ Events.AnimationDidEnd = "end"
 Events.Scroll = "scroll"
 
 # Image events
-Events.ImageLoaded = "image:load"
-Events.ImageLoadError = "image:error"
-Events.ImageLoadCancelled = "image:cancelled"
+Events.ImageLoaded = "imageload"
+Events.ImageLoadError = "imageerror"
+Events.ImageLoadCancelled = "imagecancelled"
 
 # Add all gesture events
 _.extend(Events, Gestures)
--- a/framer/Events.coffee
+++ b/framer/Events.coffee
@@ -40,8 +40,8 @@ Events.AnimationDidEnd = "end"
 Events.Scroll = "scroll"
 
 # Image events
-Events.ImageLoaded = "load"
-Events.ImageLoadError = "error"
+Events.ImageLoaded = "image:load"
+Events.ImageLoadError = "image:error"
 
 # Add all gesture events
 _.extend(Events, Gestures)
--- a/framer/Layer.coffee
+++ b/framer/Layer.coffee
@@ -681,10 +681,10 @@ class exports.Layer extends BaseClass
 			# As an optimization, we will only use a loader
 			# if something is explicitly listening to the load event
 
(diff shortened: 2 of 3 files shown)


framer/Events.coffee
```diff
@@ -40,8 +40,9 @@ Events.AnimationDidEnd = "end"
 Events.Scroll = "scroll"
 
 # Image events
-Events.ImageLoaded = "load"
-Events.ImageLoadError = "error"
+Events.ImageLoaded = "image:load"
```
koenbok: Hmm we never use this : style. I would do dom style:

`imageload` etc

framer/Layer.coffee
```diff
@@ -681,17 +691,19 @@ class exports.Layer extends BaseClass
 			# As an optimization, we will only use a loader
 			# if something is explicitly listening to the load event
 
-			if @_domEventManager.listeners(Events.ImageLoaded)?.length > 0 or @_domEventManager.listeners(Events.ImageLoadError).length > 0
-				loader = new Image()
-				loader.name = imageUrl
-				loader.src = imageUrl
+			if @listeners(Events.ImageLoaded, true) or @listeners(Events.ImageLoadError, true) or @listeners(Events.ImageLoadCancelled, true)
```
koenbok: Maybe clean up the image loader after success/error/cancel?

Not sure if it matters here, but `js -> dom -> js` like you do here (js creates dom node, dom node refers to js function) is terrible for garbage collection. So I would actually definitely do a cleanup.

test/tests/LayerTest.coffee
```diff
@@ -180,6 +180,28 @@ describe "Layer", ->
 			#layer.computedStyle()["background-size"].should.equal "cover"
 			#layer.computedStyle()["background-repeat"].should.equal "no-repeat"
 
+		it "should cancel loading when setting image to null", (done) ->
```
koenbok: Also maybe test success and error handling.

framer/Events.coffee
nvh: Fixed in d75e47a308b9de45ec9a34e5252427e90d8a54ce

framer/Layer.coffee
nvh: Do you mean destroying the dom node?

test/tests/LayerTest.coffee
nvh: That's already here: https://github.com/koenbok/Framer/pull/372/files#diff-cfd025d974e2cdd9ef6b86a3bea2f84aR158

koenbok: Bueno.

framer/Layer.coffee
```diff
@@ -639,6 +639,12 @@ class exports.Layer extends BaseClass
 	##############################################################
 	## IMAGE
 
+	cleanupImageLoader: ->
```
koenbok: This should be private: `_ cleanupImageLoader `.

framer/Layer.coffee
```diff
@@ -639,6 +639,12 @@ class exports.Layer extends BaseClass
...
+		@imageEventManager?.removeAllListeners()
+		@imageEventManager = null
+		@imageLoader = null
+
```
koenbok: All these too...

framer/Layer.coffee
nvh: Fixed in https://github.com/nvh/Framer/commit/5c0fe7e5e78fa4889a8149a382311af5b83e88ba

framer/Layer.coffee
nvh: Fixed in https://github.com/nvh/Framer/commit/5c0fe7e5e78fa4889a8149a382311af5b83e88ba

PR #374, opened 2016-06-27
Sets screen size to correct size after orientation change on device

jonastreub:


eelco:
I haven’t tested the code itself yet (will do that), but a couple of questions:
- What makes it that `BackgroundLayer` needs custom handling? Is there another way to fix this issue maybe by doing the rotation handling at another ‘level’?
- With what devices / browser did you test this? I think we should at least do:
  - [ ] iPhone + Safari
  - [ ] iPhone + Chrome
  - [ ] iPad + Safari
  - [ ] iPad + Chrome
  - [ ] Android phone + Browser?
  - [ ] Android phone + Chrome

jonastreub:
Tested on all devices, iPhone, Android and iPad, Safari, Chrome and the official apps.

jonastreub:
The backgroundLayer should always fill the entire screen. As device rotation in Framer for Mac doesn't change the window size we need a separate path. We could also fire a resize event on window but I don't tink that makes it cleaner.

eelco:
Ah, I see! Maybe that’s a good comment to add?

If you make a project with the following code the behavior is a bit weird when you rotate in Framer:
```
layerA = new Layer
	width: Screen.width

layerB = new BackgroundLayer
```

I’d expect both layerA and layerB to resize, but only `layerB` now resizes, for `layerA` I still need to refresh.

jonastreub:
`Screen.width` is just a number. Your prototype is not refreshed on orientation changes. It has always been like that. The backgroundLayer is a special type of layer in that regard.

eelco:
Thanks, gotcha! Nothing more to add, LGTM 👍

PR #376, opened 2016-07-05
Feature/scientific notation

nvh:


koenbok:
I have two questions:

- These are called _a lot_. What is the performance impact of the `?` operator?
- Shouldn't we create one `sanitize` function on top if they're all the same?

nvh:
- It's an extra if-statement for every call, but we need to check if the value is undefined, because it can be. We also could do a check if it's the scientific notation, but that's even worse, I guess. You want me to create some performance tests?
- we could do a for-loop over all the used properties, or at least factor out the magic number `20`, but as it's already only one function, wrapping it doesn't make a lot of sense to me

koenbok:
- I'd look at the generated javascript and study the effect a bit so make an educated guess about perf.
- Making 20 a static sounds fine.

nvh:
Did some performance research: http://share.framerjs.com/il21078uah96/ (see the console for the results)

screen shot 2016-07-07 at 16 30 50


As we can see, toFixed is way to slow, I'm going with the `roundToZero` method, which doesn't have a lot of overhead, but solves the problem for us

nvh:
@koenbok please review

framer/LayerStyle.coffee
```diff
@@ -90,6 +90,11 @@ exports.LayerStyle =
 
 		return css.join(" ")
 
+	_roundToZero: (num) ->
```
koenbok: Can you please make this a standalone function. We sometimes want to walk through the keys of LayerStyles and we expect them all to be css properties.

framer/LayerStyle.coffee
```diff
@@ -100,24 +105,23 @@ exports.LayerStyle =
 
 		if layer._prefer2d or layer._properties.force2d
 			return exports.LayerStyle.webkitTransformForce2d(layer)
-
 		"
 		translate3d(
-			#{layer._properties.x}px,
-			#{layer._properties.y}px,
-			#{layer._properties.z}px)
+			#{@_roundToZero layer._properties.x}px,
```
koenbok: We try to use `f(arg)` over `f arg` wherever we can.

framer/LayerStyle.coffee
nvh: Fixed in 0c9c09dbf8295b28653f1c458f04649284d91212

framer/LayerStyle.coffee
nvh: Fixed in 0c9c09dbf8295b28653f1c458f04649284d91212

PR #379, opened 2016-07-18
Preloader

eelco:
Changes:
- Also preloads audio/video media
- Exposes `addImage` as an API to add (image) URLs to preload
- Default on, outside the Framer app
- Preloader shows in the device if there is one

WIP:
- Added Framer ‘branding’ in the loading indicator (intentionally ugly at the moment)
- Added API to set your own logo / layer


framer/Extras/Preloader.coffee
```diff
@@ -5,88 +5,114 @@ class Preloader extends BaseClass
...
 	@define "isReady",
-		get: -> 
+		get: ->
 			return false if not @isLoading
-			return @_imagesLoaded.length is @_images.length
+			return @_mediaLoaded.length is @_media.length
 
 	addImagesFromContext: (context) ->
-		@addImages(_.pluck(context.layers, "image"))
+		_.pluck(context.layers, "image").map @addImage
```
koenbok: We try to use `a(x)` over `a x` wherever we can.

framer/Extras/Preloader.coffee
```diff
@@ -5,88 +5,114 @@ class Preloader extends BaseClass
...
-		@_imagesLoaded = []
+		@_media = []
+		@_mediaLoaded = []
 		@_isLoading = false
-		
+
 		@timeout = 5
 
-		@context = new Context({name: "Preloader"})
+		@context = new Context({parent: Framer.CurrentContext, name: "Preloader"})
```
koenbok: Why does this need a parent context?

framer/Extras/Preloader.coffee
```diff
@@ -109,16 +135,22 @@ class Preloader extends BaseClass
 		else
 			@end()
 
-	_handleTimeout: ->
+	_handleTimeout: =>
 		return unless @isLoading
-		console.error "Timeout"
+		console.error "Preloader timeout, ending"
```
koenbok: `console.warning` is enough maybe

framer/Extras/Preloader.coffee
eelco: To draw only draw onto the context if there is one. This is nicer if you use the preloader in Studio, it will make the preloader appear on the device screen instead of flashing the device.

eelco:
@koenbok Made some additional changes that I’d like reviewed, see commits above.

koenbok:
LGTM

koenbok:
You can maybe add some comments here and there to explain (like how you start the loop).

koenbok:
Actually, I would design it differently maybe, I would _always_ start the loop but pause it if we use the preloader.

eelco:
With `Framer.Loop.delta = 0`?

PR #385, opened 2016-07-22
add `animateTo` with different argument behaviour

nvh:
This implements the first part of #378 

It's already fully functional _and_ backwards compatible, so it can already be merged in with no trouble.


koenbok:
Merge it in master w/o the warnings so we can start playing. Let's keep developing on this branch.

PR #387, opened 2016-07-26
Feature/animation refactor

nvh:
This is a full implementation of the proposal in #378


framer/Animation.coffee
```diff
@@ -33,69 +33,48 @@ evaluateRelativeProperty = (target, k, v) ->
...
-# _runningAnimations = []
-
-# Todo: this would normally be BaseClass but the properties keyword
-# is not compatible and causes problems.
 class exports.Animation extends BaseClass
 
-	constructor: (options={}) ->
-
-		options = Defaults.getDefaults "Animation", options
+	constructor: (parameters={},options={}) ->
```
koenbok: missing space

framer/Animation.coffee
```diff
@@ -33,69 +33,48 @@ evaluateRelativeProperty = (target, k, v) ->
...
+		_.defaults(options, parameters.options)
 
-		super options
+		delete parameters.options
+		@options = _.cloneDeep Defaults.getDefaults "Animation", options
+		super parameters
+		@layer = parameters.layer ? null
```
koenbok: We should throw an error if there is no layer here, and we should make the layer a readonly property.

framer/Animation.coffee
```diff
@@ -33,69 +33,48 @@ evaluateRelativeProperty = (target, k, v) ->
...
 		set: (value) ->
 			@options?.looping = value
-			if @options?.looping and @options?.layer? and !@isAnimating
+			if @options?.looping and @layer? and !@isAnimating
 				@restart()
 
 	start: =>
-		if @options.layer is null
+		if @layer is null
 			console.error "Animation: missing layer"
```
koenbok: If we check in constructor we can skip check here

framer/Layer.coffee
```diff
@@ -883,35 +885,82 @@ class exports.Layer extends BaseClass
...
 
-	animate: (options) ->
-		# console.warn "Layer.animate is deprecated: please use Layer.animateTo instead"
-		properties = options.properties
-		delete options.properties
-		@animateTo(properties, options)
-
-	animateTo: (properties,options={}) ->
-		_.defaults(options,properties.options)
+	animateToState: (stateName, options={}) ->
```
koenbok: This needs some more comments to make clear what exactly is happening.

framer/Layer.coffee
```diff
@@ -883,35 +885,82 @@ class exports.Layer extends BaseClass
...
+			if not shouldChange
+				return null
+		finished = options.completion
+		options.completion = =>
+			@_stateMachine.emit(Events.StateDidSwitch, @_stateMachine.previousName, @_stateMachine.currentName, @)
+			finished?()
+		@animate properties, options
+
+	animate: (properties,options={}) ->
+		if typeof properties == 'string'
```
koenbok: Style ' "

framer/Layer.coffee
```diff
@@ -883,35 +885,82 @@ class exports.Layer extends BaseClass
...
-		delete options.properties
-		@animateTo(properties, options)
-
-	animateTo: (properties,options={}) ->
-		_.defaults(options,properties.options)
+	animateToState: (stateName, options={}) ->
+		properties = @_stateMachine.switchTo stateName
+		if @_stateMachine.previousName is @_stateMachine.currentName
+			shouldChange = false
+			for property,value of properties
```
koenbok: Style: space

framer/Layer.coffee
```diff
@@ -883,35 +885,82 @@ class exports.Layer extends BaseClass
...
+				if @[property] != value
```
koenbok: Style: we prefer `isnt`

koenbok: http://stackoverflow.com/questions/523643/difference-between-and-in-javascript

framer/Layer.coffee
```diff
@@ -883,35 +885,82 @@ class exports.Layer extends BaseClass
...
+
+		animatableProperties = Animation.filterAnimatableProperties(properties)
+		nonAnimatableProperties = _.omit(_.clone(properties),_.keys(animatableProperties))
 
 		start = options.start
 		start ?= true
 		delete options.start
-
-		if options.instant
+		instant = options.instant ? false
```
koenbok: http://stackoverflow.com/a/10146123/56495

koenbok:
General some styling stuff:

- `'` vs `"`
- missing spaces (function arguments)
- `!=` vs `isnt` etc.

framer/Layer.coffee
nvh: It's not a ternary, but an existential operator: http://coffeescript.org/#operators

framer/Animation.coffee
nvh: No, because I'd like to be able to clone an animation and set its layer to some other layer before starting it

nvh:
This is about finished and it could be merged in, but there are 3 breaking changes:

- The `default` state is renamed to `initial`
- `layer.states.current` has been renamed to `layer.states.currentName`
- `layer.states.previous` has been renamed to `layer.states.previousName`

Especially the last two are tricky, because `layer.states.current` and `layer.states.previous` do now contain an object with the specified state properties, so if you have the code:
```coffee
if layer.states.current == "stateA"
   print "The best state ever" # Will never happen
```
The code compiles, but doesn't do what you would expect with the previous API. We could show a warning or something, but that would hinder people using the new API as intended.

What do you think?

PR #388, opened 2016-07-26
Feature/point conversion proposal

jonastreub:
With this proposal we set out to:
- simplify the existing point conversion API
- make the API public (docs + autocomplete)

**Current API**

```coffee
Utils.convertPointToContext(point, layer, rootContext=false, includeLayer=true) -> point
Utils.convertPointFromContext(point, layer, rootContext=false, includeLayer=true) -> point

```

**Proposed API**
```coffee
Canvas.convertPointToLayer(point, layer) -> point
Screen.convertPointToLayer(point, layer) -> point

Canvas.convertPointToScreen(point) -> 		point
Screen.convertPointToCanvas(point) -> 		point

layer.convertPointToCanvas(point) -> 		point
layer.convertPointToScreen(point) -> 		point
layer.convertPointToLayer(point, layer) -> 	point
```
All conversions start as a function on the instance you are converting from. Points are always the first parameter. This parameter defaults to `{x:0, y:0}`.
When you are converting to a layer this needs to be specified as the second parameter.

### DIFF
--- a/framer/Screen.coffee
+++ b/framer/Screen.coffee
@@ -24,6 +24,15 @@ class ScreenClass extends BaseClass
 	toInspect: ->
 		return "<Screen #{Utils.roundWhole(@width)}x#{Utils.roundWhole(@height)}>"
 
+	# Point Conversion
+
+	convertPointToLayer: (point, layer) =>
+		return Utils.convertPointFromContext(point, layer, false, true)
+
+	convertPointToCanvas: (point) =>
+		ctx = Framer.CurrentContext.device.context
+		return Utils.convertPointToContext(point, ctx, true, false)
+
 	# Edge Swipe
 
 	onEdgeSwipe:(cb) -> @on(Events.EdgeSwipe, cb)
--- a/framer/Screen.coffee
+++ b/framer/Screen.coffee
@@ -26,11 +26,11 @@ class ScreenClass extends BaseClass
 
 	# Point Conversion
 
-	convertPointToLayer: (point, layer) =>
+	convertPointToLayer: (point, layer) ->
 		return Utils.convertPointFromContext(point, layer, false, true)
 
-	convertPointToCanvas: (point) =>
-		ctx = Framer.CurrentContext.device.context
+	convertPointToCanvas: (point) ->
+		ctx = Framer.Device.context
 		return Utils.convertPointToContext(point, ctx, true, false)
 
 	# Edge Swipe
--- a/test/tests/LayerTest.coffee
+++ b/test/tests/LayerTest.coffee
@@ -1148,7 +1148,7 @@ describe "Layer", ->
 
 	describe "Point conversion", ->
 
-		it "should correctly convert points to Screen", ->
+		it "should correctly convert points from layer to Screen", ->
 
 			point =
 				x: 200
@@ -1158,3 +1158,20 @@ describe "Layer", ->
 			screenPoint = layer.convertPointToScreen()
 			screenPoint.x.should.equal point.x
 			screenPoint.y.should.equal point.y
+
+		it "should correctly convert points from Screen to layer", ->
+
+			point =
+				x: 300
+				y: 200
+
+			layer = new Layer point: point
+			layerPoint = Screen.convertPointToLayer({}, layer)
+			layerPoint.x.should.equal -point.x
+			layerPoint.y.should.equal -point.y
+
+
+
+
+
+
--- a/test/tests/LayerTest.coffee
+++ b/test/tests/LayerTest.coffee
@@ -1144,3 +1144,17 @@ describe "Layer", ->
 			a1.isAnimating.should.equal false
 			a2.isAnimating.should.equal false
 			a3.isAnimating.should.equal true
+
+
+	describe "Point conversion", ->
+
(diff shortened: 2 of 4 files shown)


koenbok:
This looks really good!

jordandobson:
Do we really need to say "convert"?

`layer.pointToScreen(point) ->`

or doe we need to say "point" if you feel strongly about keeping convert.

seems like this could be a bit shorter name... but it is clear... almost too clear.

jonastreub:
I like that shorter version Jordan! Probably clear enough without the 'conversion' part.

framer/Screen.coffee
```diff
@@ -24,6 +24,15 @@ class ScreenClass extends BaseClass
 	toInspect: ->
 		return "<Screen #{Utils.roundWhole(@width)}x#{Utils.roundWhole(@height)}>"
 
+	# Point Conversion
+
+	convertPointToLayer: (point, layer) =>
+		return Utils.convertPointFromContext(point, layer, false, true)
+
+	convertPointToCanvas: (point) =>
+		ctx = Framer.CurrentContext.device.context
```
tisho: Is `CurrentContext.device` guaranteed to exist?

koenbok:
I would also love a few more unit tests:

- With(out) contexts
- For advanced transforms with rotation

framer/Screen.coffee
```diff
@@ -24,6 +24,15 @@ class ScreenClass extends BaseClass
 	toInspect: ->
 		return "<Screen #{Utils.roundWhole(@width)}x#{Utils.roundWhole(@height)}>"
 
+	# Point Conversion
+
+	convertPointToLayer: (point, layer) =>
+		return Utils.convertPointFromContext(point, layer, false, true)
+
+	convertPointToCanvas: (point) =>
```
tisho: No need to rebind the function to the current context since you're not using `@` anywhere.

framer/Screen.coffee
```diff
@@ -24,6 +24,15 @@ class ScreenClass extends BaseClass
 	toInspect: ->
 		return "<Screen #{Utils.roundWhole(@width)}x#{Utils.roundWhole(@height)}>"
 
+	# Point Conversion
+
+	convertPointToLayer: (point, layer) =>
```
tisho: No need to rebind the function to the current context since you're not using `@` anywhere.

test/tests/LayerTest.coffee
```diff
@@ -1144,3 +1144,34 @@ describe "Layer", ->
 			a1.isAnimating.should.equal false
 			a2.isAnimating.should.equal false
 			a3.isAnimating.should.equal true
+
+
+	describe "Point conversion", ->
```
tisho: Would be good to see more tests covering the rest of the `convert*` methods.

test/tests/LayerTest.coffee
```diff
@@ -1144,3 +1144,34 @@ describe "Layer", ->
...
+			layer = new Layer point: point
+			layerPoint = Screen.convertPointToLayer({}, layer)
+			layerPoint.x.should.equal -point.x
+			layerPoint.y.should.equal -point.y
+
+
+
+
+
+
```
tisho: Nit: that's a lot of whitespace :)

tisho:
Looks great!

Re: naming, I think it's better to keep the verb in there, otherwise `point` ends up looking like a verb, which is just confusing.

test/tests/LayerTest.coffee
jordandobson: Should you also be testing nested layers?

jordandobson:
@tisho I suppose you could say one of the following instead...

`layer.pointForScreen()
layer.pointInScreen()
layer.pointOnScreen()
`

I think out of any of those I would like the "For" version better.

jordandobson:
@jonastreub Do you have a test for this?

`layer.convertPointToLayer(point, layer) ->  point`

I'm a bit confused on a few of these... it seems like sometimes the point arg can be optional and other times it is needed for layers... Are there times when the point can be assumed and others where it could be passed in?

framer/Screen.coffee
jonastreub: Shorter version would be `Framer.Device`. I think the device is always there. If you choose `No Device` from the device picker this still returns a so called `fullscreen` device.

test/tests/LayerTest.coffee
jonastreub: Nested layers can be tested. Canvas conversions are hard to test since these depend on the window size which depends on the testing setup.

jonastreub:
@jordandobson Not that I am aware of. Can you give an example where the point is not optional?

jonastreub:
Thanks for your feedback! Much appreciated 💯 Will give it another pass tomorrow.

framer/Screen.coffee
tisho: OK, just wanted to confirm. I get nervous when I see code that tries to access deeply-nested properties :)

jonastreub:
@jordandobson I just tried out your suggestions but I can't make the functions consistent in their naming like that. And their meaning becomes less clear. So I'll go with the lengthier proposal for now.

PR #389, opened 2016-07-28
Feature/slider improvements

jonastreub:
- keeps slider knob a circle when changing its size
- when changing the slider its frame the knob is repositioned at the right location
- when changing the slider its frame it can change its orientation from horizontal to vertical without breaking


koenbok:
Wat should we do when it's square?

```coffee
Framer.Extras.Hints.enable()

slider = new SliderComponent
	point: Align.center
	
slider.width = 200
slider.height = 200
```

benjaminnathan:
LGTM!

PR #391, opened 2016-07-31
Fix Framer.resetDefaults

nvh:
Turns out, because of a non-deep clone, in some cases the original Framer.Defaults were overwritten, and not correctly reset when calling `Framer.resetDefaults()`. 

This fixes that, and adds a special override to resetDefaults for the tests, so we can keep using a default layer size of 100x100


koenbok:
Ha, great catch and well done.

PR #392, opened 2016-08-01
Remove invalid use of __framerInstance

nvh:
__framerInstance is only available through the Inferencer in Framer Studio, so we should never use it the change other properties of Framer objects.
We can, however use it to give more sensible debug output when using `toInspect()`

This PR does both.


jonastreub:
LGTM

PR #393, opened 2016-08-01
Change midX and midY to be at x and y when frame is 0

### LINKED ISSUE #375: Initializing layer with a width and height of 0  and midX/midY values leads to unexpected behavior when animating
In the following code, the box will (unexpectedly) begin in the top left and its position will animate to 200,200. 

I'm not sure if midX and midY are supported as initialization values, but they seem to work fine when width and height don't begin at 0.  

````coffeescript

box = new Layer
	midX:200
	midY:200
	width: 0
	height: 0
	
box.animate
	properties:
		midX:200 
		midY:200
		width: 100
		height: 100
````

nvh:
Although it is technically not correct, it makes way more sense to me this way.

Also it fixes #375

### DIFF
--- a/framer/Utils.coffee
+++ b/framer/Utils.coffee
@@ -772,9 +772,9 @@ Utils.frameGetMinX = (frame) -> frame.x
 Utils.frameSetMinX = (frame, value) -> frame.x = value
 
 Utils.frameGetMidX = (frame) ->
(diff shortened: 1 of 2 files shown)


koenbok:
Makes more sense to me too.

PR #397, opened 2016-08-02
Feature/consistent scroll events

nvh:
This implements #396


koenbok:
@nvh do we need docs updates for these?

nvh:
Hmmm, I think so, let me check

nvh:
The docs aren't incorrect, so this can be safely merged in. But they could be more specific about which events fire when.

PR #399, opened 2016-08-03
Enable native Touch events again when disabling the TouchEmulator

nvh:
This implements changes discussed in #245


koenbok:
LGTM

PR #416, opened 2016-09-13
fix video layer for ios 10

gk3:
playslinline is no longer prefixed on iOS 10. this should allow it to work on iOS 10 and previous versions

i dont think i'll break anything this time @koenbok


eelco:
Thanks!

PR #418, opened 2016-09-15
Animation Paths

nvh:
This is a version of https://github.com/koenbok/Framer/pull/151, but with the changed animation API (currently on master) merged in.


test/studio/PathAnimation.framer/app.coffee
```diff
@@ -0,0 +1,25 @@
+layer = new Layer 
+	x: 100, y: 100
+	width: 50, height: 50
+	backgroundColor: '#cbf'
+	borderRadius: 25
+
+layer.animateTo
```
tisho: Needs to be `.animate` now, right?

tisho:
Aside from a single method call in the path animation example, this is good to go! Thanks for dealing with all the merge conflicts.

test/studio/PathAnimation.framer/app.coffee
nvh: Correct! Fixed in 6fdb987

PR #420, opened 2016-09-17
Animation API refactor

koenbok:
Continuing from https://github.com/koenbok/Framer/issues/378

===
## API Overview

This is a current relevant overview of how the API will be from now on. For more information about what changed, see below.

##### Returns: Animation object

- *properties or state:* either an object with property values you want to animate to, or an existing state name.
- *options:* options for the animation (see `animationOptions`).

This creates a new animation for a layer based on a state or new property values and starts it immediately. If the properties have an `options` key, it is used as options for the animation.

If there is currently an animation on the layer, it gets stopped if it is animating the same properties as the new animation. Simply put, no two animations can be ran on width at the same time, but you can have two separate animations, one running on width and the other on height.

##### Preferred Examples

Simple examples for animating of properties.

```coffee
# Properties animation
layer.animate
	x: 200

# Properties spring animation
layer.animate
	x: 200
	options:
		curve: "spring"
```

Simple examples for animating with states.

```coffee
layer.states.test =
	x: 200

# State animation
layer.animate "test"

# State animation with options
layer.animate "test",
	curve: "spring"

```

### `layer.animateStop()`

Stop all animations for this layer.

### `layer.animationOptions`

Animation objects manage animations that target a layer and properties. An animation will tween between a start and end value, with a <a href="#animation.curves">curve</a>. The start value is determined when the animation <a href="#animation.start">starts</a>, and the end value is defined by properties. If the start values equal to end values, they won't animate.

- `curve` &mdash; A string, set to ease by default. <span>(Optional)</span>
- `curveOptions` &mdash; An object with the options of the set curve. <span>(Optional)</span>
- `time` &mdash; A number, the duration in seconds. <span>(Optional)</span>
- `delay` &mdash; A number, the delay of the animation. <span>(Optional)</span>
- `repeat` &mdash; A number, the amount of times it repeats. <span>(Optional)</span>
- `colorModel` &mdash; A string, the model to <a href="#color.models">animate colors</a> in. <span>(Optional)</span>
- `instant` &mdash; A boolean, don't animate. <span>(Optional)</span>

### `layer.states` (object)

The states object on a layer holds all the different states for a layer. They can be used to animate to, switch to, or cycled through. States are a great way to organize different visual properties of layers.

The properties of a state can be anything on a layer like `x`, `y`, etc. Properties that cannot be animated like `html` or `visible` can still be used but will not be animated. They will get set at the end of a transition.

States can have a special `options` key to hold animation options like `curve`. They will be used whenever layers get animated to that state.

There are two special states: `current` and `previous`. They refer to the current and last state that a layer is in. Additionally, the have a `name` key so you can check the current state name with `layer.states.current.name`.

#### Examples

```coffee
# Set a single state
layer.states.stateA =
	x: 100

# Animate to the state	
layer.animate("stateA")

# Go to the state without animation
layer.stateSwitch("stateA")
```

Here is a common example where we set multiple states and cycle through them.

```coffee
# Set multiple states at once
layer.states =
	stateA:
		x: 100
	stateB:
		x: 200

# On a click, go back and forth between states.
layer.onTap -> layer.stateCycle(["stateA", "stateB"])
```

### Animation

Change a property.

```coffee
layer = new Layer


## After

layer.animate
  x: 100

## Before

layer.animate
  properties:
    x: 100

```

Change a property with options.

```coffee
## Before

layer.animate
  properties:
    x: 100
  time: 0.5
  
## After

layer.animate
  x: 100
  options:
    time: 0.5
```

Change the animation curve.

```coffee
## Before

layer.animate
  properties:
    x: 100
  curve: "spring(250, 50, 0)"
  
## After

layer.animate
  x: 100
  options:
    curve: "spring(250, 50, 0)"
```

### States

Add a single state.

```coffee
## Before

layer.states.add
  stateA:
    x: 100

## After

layer.states.stateA = 
  x: 100
```

Define multiple states at once.

```coffee
## Before

layer.states.add
  stateB:
    x: 200
  stateC:
    x: 400
    
## After

layer.states =
  stateB:
    x: 200
  stateC:
    x: 400
```
Notice the subtle difference between calling a function and setting a property. This means that where previously it was possible to add multiple states multiple times, in the new API we will override the existing states when calling `layer.states = ...` again. However, this is really unlikely and one could still achieve this by doing:

```coffee
layer.states =
  stateA:
    x: 100
  stateB:
    x: 200

layer.states = _.extend layer.states,
  stateC:
    x: 300
  stateD:
    x: 400

```
Animate to state.

```coffee
## Before

layer.states.switch "stateA"

## After

layer.animate "stateA"
```

Animate to state with options.

```coffee
## Before

layer.states.add
  stateE:
    x: 200

layer.states.switch "stateE",
  curve: "ease-in"
    
## After

layer.states.stateE =
    x: 200

layer.animate "stateE",
	curve: "ease-in"

```

States can now also include optionsin the state itself. This is handy for declaring how to animate to a state as well. This has the exact same result as above.

```coffee
layer.states.stateE =
    x: 200
    options:
    	curve: "ease-in"
    	
layer.animate "stateE"
```

### Instantly switch to a state

Switching instantly will become an option of the animation

```coffee
## Before

layer.states.switchInstant "stateB"

## After

layer.stateSwitch "stateB"

## Which will be a shorthand for:

layer.animate "stateB",
    instant: true
```

This means it can also be defined directly in a state itself:

```coffee
layer.states = 
   stateA:
     x: 100
     options:
         instant: true
```

### Cycling through state

Next has been renamed to cycling, because next was pretty ambigious depending on the context.

```coffee
## Before

layer.states.next()
layer.states.next("stateB","stateC")

## After

layer.stateCycle()
layer.stateCycle(["stateB", "stateC"])  # Preferred
layer.stateCycle("stateB", "stateC")    # Also valid

# With options
layer.stateCycle ["stateB", "stateC"],
    time: 0.5
```
Notice how we use an array of states names here, to support animation options as second argument

### Special states

There are three special states that will be set automatically and can't be overridden:

- `layer.states.default` - The state the layer had upon creation.
- `layer.states.previous` - The previous state the layer was in
- `layer.states.current` - The current state the layer is in

These states contain the actual values and not (as is the case with `layer.states.current` now) the state string. The name of the previous an current states will still be available through `layer.states.previous.name`  and `layer.states.current.name`. The `name` key only gets added in the context of these special states, so the special state objects are not the same as the actual state object. They are completely equal, with the name key added.

Notice the absence of `layer.states.next`, this functionality will be provided by `layer.stateCycle()` as described above. 

### Listing all the states

We will add a new property `layer.stateNames` that lists all the names of states currently defined on a layer. This list will contain `layer.states.initial`, but won't contain `default` and `current`.

```coffee
layer.states = 
    left:
        x: Align.left
    right:
        x: Align.right

layer.animateToNextState()

print layer.states.current.name # "left"
print layer.stateNames # ["default", "left", "right"]
```

The states on a layer is pretty much a normal object. So you can use different ways to loop through the states.

```coffee
layer.states = 
    left:
        x: Align.left
    right:
        x: Align.right

# All thsese print ["default", "left", "right"]
print layer.stateNames # ["default", "left", "right"]
print Object.keys(layer.states)
print _.keys(layer.states)
print (k for k of layer.states)

# But beware of this one, it is empty: []
print (k for k in layer.states)
```
===

### Older stuff:

Some refactoring for the new animation api

- Move state switching responsibility back to the `LayerStateMachine`.
- Clean up animation events; `AnimationStart`, `AnimationStop` and `AnimationEnd`.
- Deprecate `StateWillSwitch` and `StateDidSwitch` in favor of `StateSwitchStart`, `StateSwitchStop` and `StateSwitchEnd`.
- Refactored `LayerStates` to use prototype properties, so we don't make them in constructor.

#### Changes

- You cannot override the deprecated names anymore (like `.add`) to use as a state name, because they would not show up in state names.
- Remove the `initial` state and keep `default` as is.

#### Worries

- Ideally `Object.keys(layer.states)` should be the same as `for k in layer.states` but that does not seem to be possible.
- Maybe we need to change `layer.states.current` back from the name to the actual properties and add a name if you return them so you can do `layer.states.current.name`. This would lose comparison by reference, though.
- There is a (double) circular dependency between `Layer` → `LayerStates` → `LayerStateMachine` and back. Not the end of the world, but I'm not sure if the GC is happy about this.

See: https://github.com/motif/company/issues/2633, https://github.com/motif/company/issues/2635


koenbok:
This is ready for review @nvh.

koenbok:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

> The Object.keys() method returns an array of a given object's own enumerable properties, in the same order as that provided by a for...in loop (the difference being that a for-in loop enumerates properties in the prototype chain as well).

koenbok:
So, this begs the question; what should this do?

(I think they always need to be equal)

```coffee
layer = new Layer

print "layer.stateNames", layer.stateNames
print "Object.keys", Object.keys(layer.states)
print "for..in", (k for k in layer.states)
```

The current output is:

```
» "layer.stateNames", ["default"]
» "Object.keys", []
» "for..in", []
```

koenbok:
All right, so we can't get `for..in` to behave exactly like `Object.keys()` which is sad. See the full lookup table of how they work [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties). But `for..of` works well, which is closest to what we can get.

I also have `default` replaced with `initial` _except_ when you are using deprecated methods. The only **breaking** exception is when you listen to `StateDidSwitch`, the previous/current will always be `initial`, never `default`.

koenbok:
This is ready for review @nvh

nvh:
@koenbok about default/initial: have you seen: https://github.com/motif/company/issues/2624

koenbok:
Yeah :-/ let's chat about it tomorrow.

koenbok:
Ok, re: https://github.com/motif/company/issues/2624 we decided to rip out the `initial` state for now and keep `default` the way it is.

framer/Animation.coffee
```diff
@@ -35,26 +35,58 @@ evaluateRelativeProperty = (target, k, v) ->
...
+		# Actual current api
+		if arguments.length is 3
+			layer = args[0]
+			properties = args[1]
+			options = args[2]
+
+		# Mix of current and old api
+		if arguments.length is 2
+			layer = args[0]
+			if args[1].properties
```
nvh: Shouldn't we check for null with
`if args[1].properties?`

framer/Animation.coffee
```diff
@@ -35,26 +35,58 @@ evaluateRelativeProperty = (target, k, v) ->
...
+				properties = args[1].properties
+			else
+				properties = args[1]
+			options = args[1].options if args[1].options
```
nvh: Same here `args[i].options?`

framer/Animation.coffee
```diff
@@ -35,26 +35,58 @@ evaluateRelativeProperty = (target, k, v) ->
...
+			if args[0].options
+				options = args[0].options
+			else
+				options = args[0]
+
+		delete options.layer
+		delete options.properties
+		delete options.options
+
+		# print "Animation", layer, properties, options
```
nvh: This can be removed

framer/Animation.coffee
```diff
@@ -64,6 +96,9 @@ class exports.Animation extends BaseClass
 		@_originalState = @_currentState()
 		@_repeatCounter = @options.repeat
 
+	@define "layer",
+		get: -> @_layer
```
nvh: I use later setting of layers in the animation preview of autocode. Are you ok with just using private api there?

framer/Extras/TouchEmulator.coffee
```diff
@@ -195,6 +195,7 @@ class TouchEmulator extends BaseClass
 				curve: "ease-out"
 
 	hideTouchCursor: ->
+		return unless @touchPointLayer.opacity > 0
```
nvh: Why is this in this PR?

framer/LayerStateMachine.coffee
```diff
@@ -1,51 +1,103 @@
...
+
+	@define "currentName",
+		get: -> @_currentName
 
 	@define "previousName",
 		get: -> _.last(@_previousNames)
 
 	@define "stateNames",
-		get: -> _.keys(@states)
+		get: -> Object.keys(@states)
```
nvh: Why use Object.keys and not _.keys here?

framer/LayerStates.coffee
```diff
@@ -116,3 +79,79 @@ class exports.LayerStates
 		return true if v is null
 		return true if v?.constructor?.name is "Layer"
 		return false
+
+	#################################################################
+	# Backwards compatibility
```
nvh: No subclass, like you suggested earlier?

framer/Layer.coffee
```diff
@@ -888,76 +887,45 @@ class exports.Layer extends BaseClass
...
 
 	animate: (properties, options={}) ->
-		if typeof properties == "string"
-			stateName = properties
-			return @animateToState stateName, options
 
-		#Support the old properties syntax
-		if properties.properties?
-			# console.warn "Using Layer.animate with 'properties' key is deprecated: please provide properties directly and use the 'options' key to provide animation options instead"
+		# print "layer.animate", properties, options
```
nvh: Print comment can be removed

framer/Layer.coffee
```diff
@@ -888,76 +887,45 @@ class exports.Layer extends BaseClass
...
+
+		# If the properties are a string, we assume it's a state name
+		if _.isString(properties)
+			return @animateToState(properties, options)
+
+		# Support the old properties syntax, we add all properties top level and
+		# move the options into an options property.
+		if properties.hasOwnProperty("properties")
```
nvh: Why do we check `hasOwnProperty` here instead of doing `properties.properties?`

framer/Layer.coffee
```diff
@@ -888,76 +887,45 @@ class exports.Layer extends BaseClass
...
-		animation.once Events.AnimationStop, animationFinished
-		started = animation.start() if start
-		if not started
-			# If the animation didn't start (e.g. because there are no properties to animate), call the finished handler ourselves
-			animationFinished()
-		animation
+		# With the new api we treat the properties as animatable properties, and use
+		# the special options keyword for animation options.
+		if properties.hasOwnProperty("options")
+			options = _.defaults(properties.options, options)
```
nvh: I think specifying options in the function should have precedence over specifying them in a options key

framer/Layer.coffee
```diff
@@ -1017,17 +985,18 @@ class exports.Layer extends BaseClass
 		enumerable: false
 		exportable: false
 		importable: false
-		get: -> @_stateMachine.states
+		get: ->
+			@_states ?= new LayerStates(@)
+			return @_states
```
nvh: I really dislike this behaviour of capturing the state inside of a getter, leads to unexpected behaviour. Can't we add it to the creation / addition of states?

framer/Layer.coffee
```diff
@@ -1017,17 +985,18 @@ class exports.Layer extends BaseClass
...
 		set: (states) ->
-			@_stateMachine.reset()
-			for name, state of states
-				@_stateMachine.states[name] = state
+			@states.machine.reset()
+			_.extend(@states, states)
```
nvh: We could capture the current state of the layer here, we should however be aware of the setting of states individually...

test/tests/LayerStatesTest.coffee
```diff
@@ -90,12 +90,12 @@ describe "LayerStates", ->
...
 		it "should have an initial state", ->
 			layer = new Layer
-			testStates(layer, ["initial"])
+			testStates(layer, ["default"])
 
 		it "should have an extra state", ->
 			layer = new Layer
 			layer.states.test = {x: 100}
-			testStates(layer, ["initial", "test"])
+			testStates(layer, ["default", "test"])
```
nvh: We could use `initialStateName` here as well

framer/Animation.coffee
koenbok: 👌

framer/Animation.coffee
koenbok: 👌

framer/Animation.coffee
koenbok: 👌

framer/Animation.coffee
koenbok: Yeah let's do that for now

framer/Extras/TouchEmulator.coffee
koenbok: :-/

framer/LayerStateMachine.coffee
koenbok: They are both equal, and I want to be sure `Object.keys` is the exact same as doing it on the layer.states.

framer/LayerStates.coffee
koenbok: We can't because then for...in also lists `constructor` :-(

framer/Layer.coffee
koenbok: Yep.

framer/Layer.coffee
koenbok: Well, I'd like that, but:

1) There is no hook to see if a state gets created (because it's just an object key).
2) This guarantees backwards compat

nvh: Agree with 2), we could fix 1) if we ever move to initial, but am I correct that in order to drop this, we would also need to drop the backwards compatibility of `default`?

framer/Layer.coffee
```diff
@@ -884,16 +884,13 @@ class exports.Layer extends BaseClass
...
-		return @states.machine.switchTo(stateName, options)
-
 	animate: (properties, options={}) ->
 
 		# If the properties are a string, we assume it's a state name
 		if _.isString(properties)
-			return @animateToState(properties, options)
+			# Support options as an object
+			options = options.options if options.options?
+			return @states.machine.switchTo(properties, options)
```
nvh: It would be nicer to put properties in a new variable `stateName` here to make it more clear what is happening

framer/Layer.coffee
```diff
@@ -917,14 +914,14 @@ class exports.Layer extends BaseClass
...
-		@animate(properties, _.merge(options, {instant: true}))
-
-	animateToNextState: (args..., options={}) ->
+	stateCycle: (args..., options={}) ->
 		states = []
 		states = _.flatten(args) if args.length
 		@animate(@states.machine.next(states), options)
 
+	stateSwitch: (stateName, options={}) ->
+		@animate(stateName, _.defaults({}, options, {instant:true}))
```
nvh: Do we want to support `animated: true`?

framer/LayerStateMachine.coffee
```diff
@@ -21,7 +21,7 @@ class exports.LayerStateMachine extends BaseClass
 		get: -> @_currentName
 
 	@define "previousName",
-		get: -> _.last(@_previousNames)
+		get: -> _.last(@_previousNames) or "default"
```
nvh: Why do we need this?

framer/LayerStates.coffee
```diff
@@ -15,6 +15,9 @@ deprecatedWarning = (name, suggestion) ->
 	message += ", use '#{suggestion}' instead." if suggestion?
 	console.warn message
 
+namedState = (state, name) ->
+	return _.extend({}, state, {name: name})
```
nvh: Don't we want to enforce the named state by reversing `state` and `{name: name}`

test/tests/LayerStatesTest.coffee
```diff
@@ -3,6 +3,9 @@ assert = require "assert"
 
 initialStateName = "default"
 
+cleanState = (state) ->
+	return _.pickBy(state, (value, key) -> key isnt "name")
```
nvh: This could have a better name then `cleanState`, maybe `removeKeysFromState(state,keys=["name"])`?

test/tests/LayerStatesTest.coffee
```diff
@@ -158,24 +185,24 @@ describe "LayerStates", ->
 		# it "should be a no-op to change to the current state", ->
 		# 	layer = new Layer
 		# 	layer.states.stateA = {x: 100}
-		# 	layer.switchInstant "stateA"
+		# 	layer.stateSwitch "stateA"
 		# 	animation = layer.animate "stateA", time: 0.05
 		# 	assert.equal(animation, null)
```
nvh: Why is this test disabled?

koenbok:
Review @nvh?

framer/Layer.coffee
koenbok: Meh.

framer/LayerStateMachine.coffee
koenbok: I think it's better to _always_ return something here so you don't have to do a null check (if you just created a layer and ask for `layer.states.previous.x`.

framer/LayerStates.coffee
koenbok: Yes.

test/tests/LayerStatesTest.coffee
koenbok: Because the behaviour changed. We actually do always return an animation now so that the events get called and you can skip a null check.

framer/Layer.coffee
koenbok: Well ok.

koenbok:
Ok here we go.

test/tests/LayerStatesTest.coffee
nvh: Great. Then it can be removed instead of commented out

PR #425, opened 2016-09-22
Feature/animator refactor

nvh:
For the preview in the codePanel I wanted to create an `Animator` from options without an `Animation`. This PR moves some code around to facilitate exactly that.

### DIFF
--- a/framer/Animator.coffee
+++ b/framer/Animator.coffee
@@ -30,19 +30,8 @@ class exports.Animator
 				break
 		return values
 
-	@parseFunction: (str) ->
-		result = {name: "", args: []}
-
-		if _.endsWith str, ")"
-			result.name = str.split("(")[0]
-			result.args = str.split("(")[1].split(",").map (a) -> _.trim(_.trimEnd(a, ")"))
-		else
-			result.name = str
-
-		return result
-
 	@classForCurve: (curve) ->
-		parsedCurve = @parseFunction(curve)
+		parsedCurve = Utils.parseFunction(curve)
 		animatorClassName = parsedCurve.name.toLowerCase()
 
 		if AnimatorClasses.hasOwnProperty(animatorClassName)
@@ -56,7 +45,7 @@ class exports.Animator
 	@curveOptionsFor: (options={}) ->
 		curveOptions = options.curveOptions ? {}
 		animatorClass = @classForCurve(options.curve)
-		parsedCurve = @parseFunction options.curve
+		parsedCurve = Utils.parseFunction options.curve
 		animatorClassName = parsedCurve.name.toLowerCase()
 
 		# This is for compatibility with the direct Animation.time argument. This should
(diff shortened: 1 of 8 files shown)


framer/Animator.coffee
```diff
@@ -30,6 +30,82 @@ class exports.Animator
...
 	# stop: -> AnimationLoop.remove @
+
+{LinearAnimator} = require "./Animators/LinearAnimator"
+{BezierCurveAnimator} = require "./Animators/BezierCurveAnimator"
+{SpringRK4Animator} = require "./Animators/SpringRK4Animator"
+{SpringDHOAnimator} = require "./Animators/SpringDHOAnimator"
+AnimatorClasses["linear"] = LinearAnimator
+AnimatorClasses["bezier-curve"] = BezierCurveAnimator
+AnimatorClasses["spring-rk4"] = SpringRK4Animator
+AnimatorClasses["spring-dho"] = SpringDHOAnimator
```
eelco: I liked the old syntax for this more :)

framer/Utils.coffee
```diff
@@ -76,6 +77,12 @@ Utils.nearestIncrement = (x, increment) ->
 window.requestAnimationFrame ?= window.webkitRequestAnimationFrame
 window.requestAnimationFrame ?= (f) -> Utils.delay 1/60, f
 
+Utils.animatorFor = (options) ->
```
eelco: Does this need to be exposed? Seems like an internal thing mostly, maybe mark it as such in some way?

eelco: Also, the structure seems pretty similar to _createAnimator of Animation. Can something be reused here? (E.g., optional default class?)

framer/Utils.coffee
```diff
@@ -497,16 +504,7 @@ Utils.modulate = (value, rangeA, rangeB, limit=false) ->
...
-	result = {name: "", args: []}
-
-	if _.endsWith str, ")"
-		result.name = str.split("(")[0]
-		result.args = str.split("(")[1].split(",").map (a) -> _.trim(_.trimEnd(a, ")"))
-	else
-		result.name = str
-
-	return result
+	Animator.parseFunction(str)
```
eelco: Even though we internally only use it for Animation/Animator, I would not move this and let the Utils have as little dependencies as possible.

framer/Animator.coffee
nvh: Yeah, but because of dependencies the AnimatorClasses object has to already exist, but the classes depend on the Animator Object, so I had to do it like this :/
Any suggestions how to improve this?

framer/Utils.coffee
nvh: But also not move animatorFor then?
The problem is that I want to use `animatorFor` in my code panel code, but Animator is not public in Framer. So the choices are putting it in Utils or make Animator public. Would you go for the latter instead?

framer/Animator.coffee
eelco: Oh, of course. Maybe `_.merge`? Your call.

PR #440, opened 2016-11-04
LayerDraggable reset @_isMoving

onnlucky:
Add to scroll timeline test
Hook up ScrollComponent.isAnimating to content.draggable.isAnimating
Do not emit a move event when animation is stopping, we are not actually moving

### DIFF
--- a/framer/LayerDraggable.coffee
+++ b/framer/LayerDraggable.coffee
@@ -140,7 +140,7 @@ class exports.LayerDraggable extends BaseClass
 
 		# Only reset isMoving if this was not animating when we were clicking
 		# so we can use it to detect a click versus a drag.
-		@_isMoving = @isAnimating
+		@_isMoving = @_isAnimating
 
 		# Stop any animations influencing the position, but no others.
 		for animation in @layer.animations()
@@ -278,6 +278,9 @@ class exports.LayerDraggable extends BaseClass
 		# (which would return a stale value before the simulation had finished one tick)
 		# and because @_start currently calls calculateVelocity().
 		@_isDragging = false
+
+		# reset isMoving if not animating, otherwise animation start/stop will reset it
+		@_isMoving = @_isAnimating
 		@_ignoreUpdateLayerPosition = true
 		@_lastEvent = null
 		@_eventBuffer.reset()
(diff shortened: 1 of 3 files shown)


framer/LayerDraggable.coffee
```diff
@@ -570,11 +573,13 @@ class exports.LayerDraggable extends BaseClass
...
 	_stopSimulation: =>
 		@_isAnimating = false
+		@_isMoving = false
+
 		return unless @_simulation
 		@_simulation?.x.stop()
 		@_simulation?.y.stop()
 		@_simulation = null
-		@emit(Events.Move, @layer.point)
+		#@emit(Events.Move, @layer.point)
```
eelco: Did some spelunking, this `emit` was introduced here: 8ccf6c65fa5b0744c8b0a399a30481071d4f1df0. If I understand the intent correctly, I think this is to make sure that if you use an `onMove` event listener, this code guarantees that you always get the final state also. The change was made >18 months though, so things might have changed, but I haven’t tested or reviewed code to make sure that it’s not still necessary.

onnlucky: The final state is also reached in the animator, the last time it fires will be without movement. Before this change, it fires two move events with no change and no direction, now just one.

But when we fire this one, @isMoving and @isAnimating is already false; all other move events have @isMoving and either @isAnimating or @isDragging. Except perhaps when explicitly calling stop() on an animator.

We can put back this last emit, it is the safer option.

onnlucky: ok, I moved the move event emit up, so isMoving is still true during the event

framer/LayerDraggable.coffee
```diff
@@ -278,6 +278,9 @@ class exports.LayerDraggable extends BaseClass
 		# (which would return a stale value before the simulation had finished one tick)
 		# and because @_start currently calls calculateVelocity().
 		@_isDragging = false
+
+		# reset isMoving if not animating, otherwise animation start/stop will reset it
+		@_isMoving = @_isAnimating
```
nvh: Shouldn't we use `@isAnimating` here?

framer/LayerDraggable.coffee
```diff
@@ -569,12 +572,14 @@ class exports.LayerDraggable extends BaseClass
 		@emit(Events.DragAnimationStart)
 
 	_stopSimulation: =>
+		@emit(Events.Move, @layer.point) if @_isMoving
```
nvh: Shouldn't this be `@isMoving`?

framer/LayerDraggable.coffee
onnlucky: Why go through the getter if it returns `@_isAnimating || false`? especially when manipulating a similar private property?

framer/LayerDraggable.coffee
onnlucky: idem to above, but here it is the very same property as we are manipulating a little further down ... so here we know exactly what `_isMoving` is supposed to be

PR #441, opened 2016-11-04
changed importable to true for draggable and pinchable (sven + edwin)

edwinvanrijkom:
https://github.com/motif/company/issues/2107

Also changed `pinchable` to `true`. Seems to suffer from same problem as `draggable`.


nvh:
This could use some tests.

PR #443, opened 2016-11-08
AnimationBreak

koenbok:
Proposal to add a new animation event. Currently we have:

- `AnimationStart` the animation started.
- `AnimationStop` the animation stopped at any point (always thrown).
- `AnimationEnd` the animation stopped and fully completed.

What we are missing is an event that emits when an animation is stopped mid-way (so stopped, but not ended). I propose to add `AnimationBreak`.

### DIFF
--- a/test/tests/LayerAnimationTest.coffee
+++ b/test/tests/LayerAnimationTest.coffee
@@ -698,6 +698,16 @@ describe "LayerAnimation", ->
 							layer.x.should.eql 0
 							done()
 
+			it "should call break", (done) ->
+				layer = new Layer
+				layer.animate
+					x: 100
+					options:
+						time: 0.1
+						onBreak: ->
+							done()
+				layer.animateStop()
+
 			it "should call stop", (done) ->
 				layer = new Layer
 				layer.animate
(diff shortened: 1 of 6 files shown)


nvh:
I would name it `AnimationCancel`

jordandobson:
Yeah break sounds like it was going out for coffee real quick and will be
right back.

I was thinking "Halted" or "Halt" as that implies some progress was made.

Cancel feels like it was abandoned and no progress would be made maybe?

koenbok:
@nvh halt or cancel?

test/tests/LayerAnimationTest.coffee
```diff
@@ -698,6 +698,16 @@ describe "LayerAnimation", ->
 							layer.x.should.eql 0
 							done()
 
+			it "should call break", (done) ->
+				layer = new Layer
+				layer.animate
+					x: 100
+					options:
+						time: 0.1
```
nvh: There's a `AnimationTime` variable for this

nvh:
I agree a bit with Jordan that it would be nice to indicate progress, but as we discussed previously, an animation can be stopped before it begins (when there's a delay). So when cancel is emitted, progress doesn't necessarily have been made.

koenbok:
Let's go with @jordandobson. Halt it is.

jordandobson:
I'm excited to see this... This is great. I could have used this a few times to address when animations never completely finished between views so that I could resolve any issues.

PR #444, opened 2016-11-14
Fix passing any shape as parameter to SVGLayer

gusso:
Considering `type` was passed but not used, I suppose this was the intention.


gusso:
Fixed on 9976bf3

PR #446, opened 2016-11-23
when using mouse wheel, the scroll.direction was bogus

onnlucky:
by temporarily setting _lastEvent to draggable, the direction is correctly
calculated when there is no velicity

### DIFF
--- a/framer/Components/ScrollComponent.coffee
+++ b/framer/Components/ScrollComponent.coffee
@@ -401,9 +401,13 @@ class exports.ScrollComponent extends Layer
 
 		@content.point = point
 
+		# make sure _lastEvent is set on our draggable, and that it has a delta.{x,y}
+		event.delta = {x: deltaX, y: deltaY}
+		@content.draggable._lastEvent = event
 		@content.emit(Events.Move, point)
 		@emit(Events.Scroll, event)
 		@_onMouseWheelEnd(event)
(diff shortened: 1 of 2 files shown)


eelco:
This behavior is still a bit odd, I think:

```
scroll = new ScrollComponent
	width: Screen.width
	height: 100
	mouseWheelEnabled: true
	scrollHorizontal: false

layerA = new Layer
	image: Utils.randomImage()
	parent: scroll.content
		
scroll.onScroll ->
	print "scroll", @.direction
	print "draggable", @.draggable.direction
```

Using a trackpad, if you scroll by moving two fingers up and down (only works when `mouseWheelEnabled`, the scroll direction will be fine, but the draggable direction is `null`. When you scroll the content by dragging, the scroll direction is `undefined`, but the draggable direction is correct.

I think the `null` value is actually correct, because you’re not dragging the layer, but scrolling it. I would expect the scroll direction to always have a value though.

onnlucky:
not that before, these were simply always returning "down" ... 

when doing this, it is slightly better
```
scroll.onMove ->
	print "scroll", @.direction
	print "draggable", @.draggable.direction
```
but I will see about fixing this

eelco:
Yes, sorry for highjacking the pull request :) I think this works well and should be merged and the point I brought up is a ‘new’ issue.

PR #436, opened 2016-10-26
Layer selector

CAPSLOCKUSER:
Some updates to this PR.

https://github.com/koenbok/Framer/pull/336

PR #456, opened 2016-12-15
Feature/layer selector

nvh:
This is building upon #336 and #436, changing:
- The names of some of the methods
- The implementation of the methods to be more defined in terms of each other

It still has supports these selectors:

| Selector | Result |
| --- | --- |
| A | Any layer named A |
| A B | Any layer named B that is a descendant of a layer named A (that is: a child, or a child of a child, etc.) |
| A > B | Any layer named B that is a child (i.e. direct child) of a layer named A |
| A, B | Any layer named A or any layer named B |
| * | Any layer (wildcard character) |

And adds a couple of API's:
```coffee
layer = new Layer

# Selects layer(s) that are decendants of this layer
layer.selectChild("A")
layer.selectAllChildren("A*")

context = new Framer.Context(name: "Test")
# Selects layer(s) in a specific context
context.selectLayer("A")
context.selectAllLayers( "A*")

# Selects layer(s) in the current context
Layer.select("A") 
Layer.selectAll("A*")

```

### DIFF
--- a/framer/Context.coffee
+++ b/framer/Context.coffee
@@ -150,8 +150,11 @@ class exports.Context extends BaseClass
 		return layer if layer
 		return @layerForElement(element.parentNode)
 
-	selectLayers: (selector) ->
-		return _.filter @_layers, (layer) -> Utils.layerMatchesSelector(layer, selector)
+	selectLayer: (selector) ->
+		Utils.findLayer(@_layers, selector)
+
+	selectAllLayers: (selector) ->
+		Utils.filterLayers(@_layers, selector)
 
 	layout: =>
 		@rootLayers.map (l) -> l.layout()
--- a/framer/Context.coffee
+++ b/framer/Context.coffee
@@ -150,6 +150,9 @@ class exports.Context extends BaseClass
 		return layer if layer
 		return @layerForElement(element.parentNode)
 
+	selectLayers: (selector) ->
+		return _.filter @_layers, (layer) -> Utils.layerMatchesSelector(layer, selector)
+
 	layout: =>
 		@rootLayers.map (l) -> l.layout()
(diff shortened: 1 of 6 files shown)


awt2542:
Really excited about this. A global shortcut variable would be great as well. I've experimented a bit with ƒ() but it's hard to write on mobile and conflicts with Atom's default key bindings. Still – something like $() would be interesting since it in theory could replace layer names as exposed variables and avoid issues like protected names, illegal characters, multiple layers sharing the same name etc.

nvh:
A global shortcut variable doesn't feel like it fits the Framer Library to me, but you could easily create it yourself by adding
```coffee
$ = Layer.selectAll
```
to your project

nvh:
Also: beware that only works with explicit layer names, set using the `name` property of a layer (which is the case for imported layers). In Framer we add some magic to show sensible names to the layer panel so `button = new Layer` shows up as "button" in the layer panel, although it doesn't have an explicit name.

marckrenn:
I think the differentiation between `Layer.select` and `Layer.selectAll` may cause more trouble through inconsistency than it actually helps.

For example, if your projects has multiple layers with the exact same name (can easily happen with a generic name like "button"), which layer does `Layer.select` pick?

Hence I'd propose to rewrite the `Layer.select` method in a way that it always returns an array of matching layers and get rid of `Layer.selectAll` completely.

declanv:
First off...THANK YOU. This functionality is a phenomenal addition. Might I humbly suggest a :not selector, similar to the [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/:not) or [jQuery](http://api.jquery.com/not/) selectors in functionality?

PR #458, opened 2016-12-23
add reset methods to TouchEmulator and GestureInputRecognizer

onnlucky:
can be used for webviews that mght miss mouseout/mouseup events because
they are hidden or disabled from the outside


nvh:
LGTM!

PR #459, opened 2016-12-28
Fix/slider broken anims

onnlucky:
If animations result in errors during update, animation loop will stop ticking, forever. Even after a context reset. Fix that by first requesting a next frame, then doing logic.

Such broken animations could be triggered by calling SliderComponent.animateToValue(); so adding guards against weird slider values. (There is can be a loop in between knop position change -> update value -> change knob position, normally broken by value == lastvalue, but !(NaN == NaN))

### DIFF
--- a/framer/AnimationLoop.coffee
+++ b/framer/AnimationLoop.coffee
@@ -56,12 +56,12 @@ class exports.AnimationLoop extends EventEmitter
 		tick = (timestamp) ->
 
 			if animationLoop.raf
-				update()
 				window.requestAnimationFrame(tick)
+				update()
 			else
 				window.setTimeout ->
-					update()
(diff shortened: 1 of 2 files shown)


nvh:
Is it safe to call requestAnimationFrame before doing update? Can't this result in two ticks happening before the update has taken place, causing race conditions?

onnlucky:
Not unless after update() it can result in two ticks ... I mean usually update() is practically a no-op when the lists are empty.

It is defined as called before the *next* repaint. So not the current repaint.

nvh:
But what if our `update()` takes 2 repaints to complete? Usually we would drop a frame, but now we already requested to be called again causing a second call to `update()` while it is still not finished with the previous pass, right?

PR #462, opened 2017-01-03
Add graphics layers

koenbok:
Add `SVGLayer`, `CanvasLayer`.

#### SVGLayer

- `.svg: SVGSVGElement ` the svg element in the dom.
- `.addShape(shape: string): SVGXXXElement` create a new SVG element.

#### CanvasLayer

We automatically correct the canvas for retina.

- `.canvas: HTMLCanvasElement` the canvas element in the dom.
- `.ctx: CanvasRenderingContext2D` the canvas 2d context.

### DIFF
--- a/framer/CanvasLayer.coffee
+++ b/framer/CanvasLayer.coffee
@@ -1,0 +1,29 @@
+{_} = require "./Underscore"
+
+{Layer} = require "./Layer"
+
+DPR = window.devicePixelRatio or 1
+
+class exports.CanvasLayer extends Layer
+	
+	constructor: (options={}) ->
+
+		super _.defaults options,
+			backgroundColor: null
+		
+		@_canvas = document.createElement("canvas")
+		@_ctx = @canvas.getContext("2d")
+		
+		@_update()
+		@_element.appendChild(@canvas)
+		@on("update:frame", @_update)
+	
+	@define "canvas", get: -> @_canvas
+	@define "ctx", get: -> @_ctx
+
+	_update: =>
+		@canvas.width = @width * DPR
+		@canvas.height = @height * DPR
+		@canvas.style.width = @width + "px"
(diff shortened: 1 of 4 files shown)


marckrenn:
Added two features to CanvasLayer, which might miss the point but I'll share them anyway. Since I don't see this being used, I didn't bother with PR'ing this branch, so here's the [Framer Cloud link](https://framer.cloud/sFuJW/) instead.

**Added features:**
1.`layer.image` per default draws local images to the layer's canvas, instead of applying it to the layer itself.
2. `layer.pixel[x][y]`, a convenient way to access the color value of a certain pixel of the canvas. Returns a _Color_-object.

**ToDo:**
- Provide onPixelsReady-method

**CanvasLayer feature requests:**
- Methods: Draw pixels / lines / rectangles / ellipses to canvas (parity with SVGLayer)
- Use backgroundColor to fill the canvas, instead of applying it to the layer's background (Why would anyone want to do that?)

**CanvasLayer feedback:**
- In my opinion, CanvasLayers' backgroundColor should be set to _null_ only after a canvas color or canvas image was set. Until then the default backgroundColor should be used.

framer/CanvasLayer.coffee
```diff
@@ -0,0 +1,29 @@
...
+
+		super _.defaults options,
+			backgroundColor: null
+		
+		@_canvas = document.createElement("canvas")
+		@_ctx = @canvas.getContext("2d")
+		
+		@_update()
+		@_element.appendChild(@canvas)
+		@on("update:frame", @_update)
```
marckrenn: You sure this shouldn't be `@on("change:frame", @_update)`?
At least this change (no pun) gives me the result I was expecting.

marckrenn: refering to https://framer.cloud/QqxrA/

PR #467, opened 2017-02-10
Feature/spring improvements

nvh:


framer/Animators/Curves.coffee
```diff
@@ -0,0 +1,132 @@
...
+Spring.computeDuration = computeDuration
+Spring.computeDampingRatio = computeDampingRatio
+
+
+exports.Spring = Spring
+exports.Bezier = Bezier
+parseFunction = (string) ->
+	return null unless _.isString string
+
+	regex = /.*(Spring|Bezier)(?:\(\s*{?([a-zA-Z\d:\s,.]*)}?\s*\)|\.(\w+))?/
```
eelco: Why `a-zA-Z` instead of `\w`?

This regex can use a comment with a bit of explanation of what it tries to achieve.

framer/Animators/Curves.coffee
```diff
@@ -0,0 +1,132 @@
...
+		when "ease-in" then Bezier.easeIn
+		when "ease-out" then Bezier.easeOut
+		when "ease-in-out" then Bezier.easeInOut
+		when "bezier-curve", "cubic-bezier"
+			Bezier(args...)
+		when "spring", "spring-rk4", "spring-tfv"
+			pairs = _.zipWith(["tension", "friction", "velocity", "tolerance"], args, [250, 25, 0, 1 / 100], (key, value, defaults) -> [key, value ? defaults])
+			object = _.fromPairs(pairs)
+			Spring(object)
+		when "spring-dd"
```
eelco: What’s `spring-dd`?

Also, I think we need a nice error/exception when somebody uses the old `spring-dho`

PR #469, opened 2017-02-10
Some MS Edge fixes

adrientetar:
r? @nvh

### DIFF
--- a/framer/Print.coffee
+++ b/framer/Print.coffee
@@ -23,7 +23,7 @@ class Printer
 			@_printLayer.ignoreEvents = false
 			@_printLayer.html = ""
 			@_printLayer.style =
-				"font": "12px/1.35em Menlo"
+				"font": "12px/1.35em Menlo, Consolas, monospace"
 				"color": "rgba(0, 0, 0, .7)"
 				"padding": "8px"
 				"padding-bottom": "30px"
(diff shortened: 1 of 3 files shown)


adrientetar:
These fixes should be pretty uncontroversial...

r? @koenbok

framer/Print.coffee
```diff
@@ -35,7 +35,7 @@ class Printer
 
 			@_closeButton = new Layer
 				parent: @_container
-				html: '<svg><g stroke="#B8B8B8"><path d="M1,1 L8,8"></path><path d="M1,8 L8,1"></path></g></svg>'
+				html: '<svg viewBox="0 0 9 9"><g stroke="#B8B8B8"><path d="M1,1 L8,8"></path><path d="M1,8 L8,1"></path></g></svg>'
```
nvh: This misplaces the icon in Safari and Chrome:

screen shot 2017-02-21 at 17 09 41

adrientetar: Thanks. I reverted this change and also noticed I see the same sizing issue in Chrome (using the Project template in the repo), the small close button sizes to 300x150!

svgwbug

Couldn't figure out why, but setting **overflow: hidden** on the body tag seems like a workaround.

PR #476, opened 2017-03-03
Small tweaks to fix certain broken Layer image URLs

mmccurdy:
I was working on a prototype that required assigning `Layer` image URLs of the following form:

`https://foo.bar.com/photo/transcode?url=http%3A%2F%2F127.0.0.1%2Fthumb%2F1484295501&width=132&height=200`

... and I was running in to two semi-related issues addressed by this PR.

First, URLs of that form were incorrectly being identified as local because they contain the string `127.0.0.1` even though it's buried deep in the query string. 	3b6c061 addresses this by applying stricter matching (while copying the still-somewhat-imprecise scheme matching `[a-zA-Z]{1,8}:\/\/` from the method above).

Second, when the `Layer` evaluates whether to include the `nocache` param, it doesn't take into account whether a querystring is already present, and just appends `?nocache=` in all cases, leading to a malformed request for the (incorrectly identified-as-local) URL form above. f9c6611 addresses this by looking for a `?` and appending with `&` instead if present.

I realize this is probably a niche case, but I believe each of these is low-risk and more correct in its own right, so hopefully worth consideration. Thanks!

### DIFF
--- a/framer/Layer.coffee
+++ b/framer/Layer.coffee
@@ -714,7 +714,8 @@ class exports.Layer extends BaseClass
 			# imageUrl = Config.baseUrl + imageUrl
 
 			if @_alwaysUseImageCache is false and Utils.isLocalAssetUrl(imageUrl)
-				imageUrl += "?nocache=#{NoCacheDateKey}"
+				imageUrl += if /\?/.test(imageUrl) then '&' else '?'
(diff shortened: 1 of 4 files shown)


nvh:
LGTM! Is it possible for you to add some tests for this?

mmccurdy:
Done!

nvh:
🎉 Thanks!

PR #477, opened 2017-03-08
Constrained Fix

benjamindenboer:


benjamindenboer:
This makes the `constrained` property a part of the constructor itself — matching our recently launched `RangeSlider` and the Auto-Complete.

benjamindenboer:
cc @nvh

PR #484, opened 2017-03-30
Making sure jp2 doesn’t get flagged as supported on Firefox

edwinvanrijkom:


framer/Utils.coffee
```diff
@@ -406,7 +409,7 @@ Utils.devicePixelRatio = ->
 	window.devicePixelRatio
 
 Utils.isJP2Supported = ->
-	return Utils.isWebKit() and not Utils.isChrome()
+	return Utils.isWebKit() and not (Utils.isChrome() or Utils.isFirefox())
```
koenbok: It will never be webkit `and` Firefox :-)

```
return false if Utils.isFirefox()
return Utils.isWebKit() and not Utils.isChrome()
```

adrientetar:
To me the isWebkit() function should be fixed to exclude Firefox, rather than hot patching a single problematic callsite. I did that for Edge already; these issues suggest that the core check in isWebkit() is flimsy, even Chrome which returns true and Safari has quite diverged now that the webkit engine was forked and is developed separately. I think the isWebkit() function should be deprecated/removed in favor of isSafari()/isChrome() etc.

PR #489, opened 2017-04-07
Added tests for layer.style.boxShadow propagation

marckrenn:
In response to [https://www.facebook.com/groups/760188533992744/permalink/1493404717337785/?comment_id=1493538680657722&reply_comment_id=1493999860611604&comment_tracking=%7B%22tn%22%3A%22R%22%7D](https://www.facebook.com/groups/760188533992744/permalink/1493404717337785/?comment_id=1493538680657722&reply_comment_id=1493999860611604&comment_tracking=%7B%22tn%22%3A%22R%22%7D)

@nvh

### DIFF
--- a/test/tests/LayerTest.coffee
+++ b/test/tests/LayerTest.coffee
@@ -584,6 +584,27 @@ describe "Layer", ->
 			layer.shadowColor = null
 			layer.style.boxShadow.should.equal "rgba(0, 0, 0, 0) 10px 10px 10px 10px"
 
+		it "should set the shadow via layer.style.boxShadow", ->
+
+			layer = new Layer
+				style: boxShadow: "inset 10px 10px 10px rgba(255, 0, 0, 0.5)"
+
+			# layer.style.boxShadow should propagate to Framer's shadow properties
+			layer.shadowX.should.equal 10
+			layer.shadowY.should.equal 10
+			layer.shadowBlur.should.equal 10
+			layer.shadowSpread.should.equal 10
+			layer.shadowColor.should.equal "rgba(255, 0, 0, 0.5)"
+
+			layer.shadowColor.r.should.equal 255
+			layer.shadowColor.g.should.equal 0
+			layer.shadowColor.b.should.equal 0
+			layer.shadowColor.a.should.equal 0.5
+
+			layer.style.boxShadow.should.equal "inset 10px 10px 10px rgba(255, 0, 0, 0.5)"		
+
+	describe "Events", ->
+			
 		it "should remove all events", ->
 			layerA = new Layer
 			handler = -> console.log "hello"


test/tests/LayerTest.coffee
```diff
@@ -584,6 +584,27 @@ describe "Layer", ->
...
+			layer.shadowColor.should.equal "rgba(255, 0, 0, 0.5)"
+
+			layer.shadowColor.r.should.equal 255
+			layer.shadowColor.g.should.equal 0
+			layer.shadowColor.b.should.equal 0
+			layer.shadowColor.a.should.equal 0.5
+
+			layer.style.boxShadow.should.equal "inset 10px 10px 10px rgba(255, 0, 0, 0.5)"		
+
+	describe "Events", ->
```
marckrenn: I think this was missing, as the tests below aren't related to the shadow properties in any way.

test/tests/LayerTest.coffee
```diff
@@ -584,6 +584,27 @@ describe "Layer", ->
 			layer.shadowColor = null
 			layer.style.boxShadow.should.equal "rgba(0, 0, 0, 0) 10px 10px 10px 10px"
 
+		it "should set the shadow via layer.style.boxShadow", ->
+
+			layer = new Layer
+				style: boxShadow: "inset 10px 10px 10px rgba(255, 0, 0, 0.5)"
```
marckrenn: 1. The declaration of style-properties in the constructor may be against your testing guidelines, but this was done on purpose, as this is what causes the problem. Setting `layer.style.boxShadow` after init works right now, but this changes aren't propagated to Framer's shadow properties, yet. See lines 593-597.

2. I've included the optional `inset`-property as well, as this is the primary reason for why people are using this hack / workaround. Support for the `inset`-property would be desired.

Potential ways to fix these problems:

- Rework [this part of Layer.coffee](https://github.com/koenbok/Framer/blob/master/framer/Layer.coffee#L49-L55). I believe there's a problem with how you guys map CSS properties to `layer.style`: `box-shadow` for instance, gets mapped to `layer.style.boxShadow` and loses the hyphen in the process. Hence why I think the conditional in the code block linked above does not work properly.

- Introduce `layer.shadowInset = true` and include it here: [LayerStyle.coffee](https://github.com/koenbok/Framer/blob/master/framer/LayerStyle.coffee#L160-L169). If you decide to do that, a test for `layer.shadowInset = true` should be added as well.

jordandobson:
I also use it to draw multiple outlines or inter lines. Sometimes a center line.

That's the nice part about using a string is you can apply multiple just like in CSS.

Getting it overridden or ignored sucks.

nvh:
Git bisect shows that this is the first commit that broke this: https://github.com/koenbok/Framer/commit/f0f2c9f2bbf084e6f8f5617e862b458b14179e50

@koenbok do you have any idea what's going on?

jordandobson:
@marckrenn - Any thoughts on being able to do multiple shadows?

All this time I thought, primarily because in the docs it's listed next to all the other CSS filters, that the shadow stuff was actually dropShadow: https://css-tricks.com/almanac/properties/f/filter/#article-header-id-10

Here's an actual use case for what I'm doing...
`templateBorder_over = "inset 0 0 0 1px #fff, inset 0 0 0 3px #ff435e, inset 0 0 0 1px #e5e6e9, inset 0 0 0 4px #fff, inset 0 0 0 5px #f6f7f8, 0 1px 4px #e5e6e9"`

jordandobson:
Maybe it makes sense to allow you to add an array of objects to something like 

```
layer.shadow = [
  {inset:true, x: 3, y: 5, blur: 1, spread: 2, color: "#fff"},
  {inset:false, x: 0, y: 0, blur: 0, spread: 2, color: "#000"}
]
```

That could mess up animation however... So I'm not entirely sure if that makes sense.

marckrenn:
I don't think they should support multiple shadows like that (using an array), for a bunch of reasons – breaking animation being one of them.

Maybe they should just revert to the previous, _working_ state? Idk

jordandobson:
Yeah I think so... I doubt we want to match what comes in from that style property and map it to something like that... especially if it would ruin the ability to do multiple shadows.

PR #494, opened 2017-04-27
MIDIComponent compatibility with framer.cloud / iFrames

marckrenn:
@eelco @nvh

Needs testing.

In reference to
https://www.facebook.com/groups/760188533992744/permalink/1518013801543543/


eelco:
Cherry picked the commit that uses the parent.

PR #511, opened 2017-06-09
Feature/mouse touch support

nvh:
Fixes https://github.com/motif/company/issues/5015

### DIFF
--- a/framer/Events.coffee
+++ b/framer/Events.coffee
@@ -25,14 +25,13 @@ Events.enableEmulatedTouchEvents = (enable=true) ->
 		Events.TouchStart = "touchstart"
 		Events.TouchEnd = "touchend"
 		Events.TouchMove = "touchmove"
+	Events.Click = Events.TouchEnd
 
 Events.enableEmulatedTouchEvents(false)
 
-# Let's make sure the touch events work on desktop too
-if not Utils.isTouch()
+# Let's make sure the touch events work with a mouse too
+if Utils.isMouse()
 	Events.enableEmulatedTouchEvents()
-
-Events.Click = Events.TouchEnd
 
 # Animation events
 Events.AnimationStart = "start"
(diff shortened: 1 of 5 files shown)


framer/Events.coffee
```diff
@@ -25,15 +25,14 @@ Events.enableEmulatedTouchEvents = (enable=true) ->
 		Events.TouchStart = "touchstart"
 		Events.TouchEnd = "touchend"
 		Events.TouchMove = "touchmove"
+	Events.Click = Events.TouchEnd
```
eelco: Tab vs Space :)

framer/Events.coffee
```diff
@@ -25,15 +25,14 @@ Events.enableEmulatedTouchEvents = (enable=true) ->
...
 
 Events.enableEmulatedTouchEvents(false)
 
-# Let's make sure the touch events work on desktop too
-if not Utils.isTouch()
+# Let's make sure the touch events work with a mouse too
+if Utils.isMouse()
```
eelco: This seems like it still should be `not Utils.isTouch()`? (only enable the emulator if there’s no touch)

nvh:
Only didn't add the `isMouse()` function to the `isTouch` checks in https://github.com/koenbok/Framer/blob/master/framer/GestureInputRecognizer.coffee, but I think that doesn't make sense

framer/Events.coffee
eelco: I now see there’s an extra guard in the enable. Maybe this `if` should be removed and make the logic for enabling live purely there?

framer/Utils.coffee
```diff
@@ -353,6 +353,11 @@ Utils.isTouch = ->
 	window.ontouchmove is null and
 	window.ontouchend is null
 
+Utils.isMouse = ->
+	window.onmousedown is null and
+	window.onmouseup is null and
+	window.onmousemove is null
+
```
eelco: Does this work? I’d expect for backwards compatibility this would still return `true` on iOS?

nvh: You're right. Stupid oversight

PR #515, opened 2017-06-12
Linear Gradients

jonastreub:
New:
• Linear Gradient class
• Can be set on the `image` property of a layer instance
• Or use the dedicated `gradient` property which maps to the `image` property
• Gradients are animatable


framer/Color.coffee
```diff
@@ -303,10 +303,10 @@ class exports.Color extends BaseClass
...
 		colorB = new Color(colorB)
 
-		return false if colorA.r isnt colorB.r
-		return false if colorA.g isnt colorB.g
-		return false if colorA.b isnt colorB.b
-		return false if colorA.a isnt colorB.a
+		return false if Math.round(colorA.r) isnt Math.round(colorB.r)
+		return false if Math.round(colorA.g) isnt Math.round(colorB.g)
+		return false if Math.round(colorA.b) isnt Math.round(colorB.b)
+		return false if Math.round(colorA.a) isnt Math.round(colorB.a)
```
nvh: This is a bit weird because now 0.8 and 1.2 are equal, but 0.4 and 0.6 aren't. I think it's better to use an equality function with a tolerance.

framer/Layer.coffee
```diff
@@ -882,6 +883,14 @@ class exports.Layer extends BaseClass
 			@_getPropertyValue "image"
 		set: (value) ->
 
+			if LinearGradient.isLinearGradient(value)
+				oldValue = @_getPropertyValue "image"
```
nvh: You can move the currentValue on line 897 below above the if statement and use that.

framer/Color.coffee
nvh: Also, this equality change needs some tests, I think

framer/LinearGradient.coffee
```diff
@@ -0,0 +1,76 @@
+{_} = require "./Underscore"
+{BaseClass} = require "./BaseClass"
+{Color} = require "./Color"
+
+class exports.LinearGradient extends BaseClass
+	constructor: (options = {}) ->
+		options.start ?= "black"
+		options.end ?= "white"
+		options.angle ?= 0
```
nvh: I think it's clearer (and more consistent with the rest of the lib) to use `_.defaults` for this

nvh:
```
_.defaults options,
   start: "black"
   end: "white"
   angle: 0
```

framer/Color.coffee
nvh: What you probably want is:
```
Math.abs(colorA.r - colorB.r) < tolerance
```

PR #517, opened 2017-06-13
Prevent rounding bug in LayerDraggable

daneden:
A colleague recently discovered a bug that caused draggable layers with constraints to incorrectly animate to a slightly different position than specified. The bug can be reproduced using the following code: https://gist.github.com/daneden/f5a0826168ea9626af3091d5b08c7911

Dragging the layer `myCard` up causes it to return to position `y: 89` instead of `y: 90`. This PR addresses the issue by using `Math.round` instead of `parseInt` to ensure the layer is returned to the correct constraint.

```js
> parseInt(8.999)
  8
> Math.round(8.999)
  9
```

### DIFF
--- a/framer/LayerDraggable.coffee
+++ b/framer/LayerDraggable.coffee
@@ -531,7 +531,7 @@ class exports.LayerDraggable extends BaseClass
 		return unless @_simulation
 
 		# Round the end position to whole pixels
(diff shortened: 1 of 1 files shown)


gk3:
cc @koenbok

nvh:
Makes sense to me

PR #524, opened 2017-06-30
Gradient Tests

eelco:
Implements and adds tests for behavior such as discussed:
- Animate sub properties of gradients
- Animate from/to `null` gradients
- Copy/value semantics when setting object values on `borderRadius`, `borderWidth` and `gradient` properties
- Ability to set subproperties of those properies directly via layer.

Because `Proxy` is ES6, I implemented a bit more minimal proxy class, looking at Proxy polyfills.

Questions:
- Does it work as expected?
- Is the code OK?
- Should the ReSetterProxy be more minimal?
- Where should the ReSetterProxy go?

### DIFF
--- a/framer/Animation.coffee
+++ b/framer/Animation.coffee
@@ -278,7 +278,7 @@ class exports.Animation extends BaseClass
 		for k, v of @_stateB
 			if Color.isColorObject(v) or Color.isColorObject(@_stateA[k])
 				@_valueUpdaters[k] = @_updateColorValue
-			else if Gradient.isGradient(v) or Gradient.isGradient(@_stateA[k])
+			else if Gradient.isGradientObject(v) or Gradient.isGradientObject(@_stateA[k])
 				@_valueUpdaters[k] = @_updateGradientValue
 			else
 				@_valueUpdaters[k] = @_updateNumberValue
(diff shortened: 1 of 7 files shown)


framer/Animation.coffee
```diff
@@ -294,8 +296,14 @@ class exports.Animation extends BaseClass
 		@_target[key] = Color.mix(@_stateA[key], @_stateB[key], value, false, @options.colorModel)
 
 	_updateGradientValue: (key, value) =>
+		if not @_stateB[key] and value is 1
+			@_target[key] = @_stateB[key]
```
nvh: Are we resetting the target value here? because `@_stateB[key]` is empty, right?

test/tests/LayerTest.coffee
```diff
@@ -582,6 +583,75 @@ describe "Layer", ->
...
+
+		it "should set sub-properties of gradients", ->
+			layer = new Layer
+				gradient:
+					start: "blue"
+			layer.gradient.end = "yellow"
+			layer.gradient.start.isEqual("blue").should.be.true
+			layer.gradient.end.isEqual("yellow").should.be.true
+			layer.style["background-image"].should.equal("linear-gradient(0deg, rgb(0, 0, 255), rgb(255, 255, 0))")
+
```
nvh: Maybe move these into separate border and gradient `describe` blocks?

framer/Layer.coffee
```diff
@@ -79,14 +79,27 @@ layerProperty = (obj, name, cssProperty, fallback, validator, transformer, optio
...
-	new Proxy value,
-		set: (proxiedValue, subProperty, subValue) ->
-			clone = Object.assign(new proxiedValue.constructor, proxiedValue)
-			clone[subProperty] = subValue
-			layer[property] = clone
-			return true
+	return value unless _.isObject(value)
+	new ReSetterProxy value, (proxiedValue, subProperty, subValue) ->
+		proxiedValue[subProperty] = subValue
+		layer[property] = proxiedValue
```
nvh: This might be obvious now, but will be a head-scratcher in a year. You might want to add some comments on why and how we are doing this.

framer/Animation.coffee
eelco: Yes, because we override `@_stateB[key]` to have “something” to animate to, we make sure that we do set the original value when the animation finishes.

PR #525, opened 2017-07-05
Make borderRadius and borderWidth animatable when using object values.

eelco:


extras/Studio.framer/framer/version
```diff
@@ -1 +1 @@
-7
\ No newline at end of file
+8
```
nvh: Is it necessary to bump the version? Isn't this backwards compatible?

test/tests/LayerAnimationTest.coffee
```diff
@@ -879,3 +879,73 @@ describe "LayerAnimation", ->
...
+
+	describe "Border animations", (done) ->
+
+		it "should animate borders from number to number", (done) ->
+			layer = new Layer
+				borderWidth: 10
+				borderRadius: 20
+			layer.on Events.AnimationEnd, ->
+				layer.borderWidth.should.equal 30
+				layer.borderRadius.should.equal 40
```
nvh: I understand the drive to combine these, but they should all be separated into tests for borderRadius and borderWidth

framer/Animation.coffee
```diff
@@ -338,7 +360,7 @@ class exports.Animation extends BaseClass
 				animatableProperties[k] = v
 			else if Color.isValidColorProperty(k, v)
 				animatableProperties[k] = new Color(v)
-			else if k is "gradient"
+			else if k in ["gradient", "borderRadius", "borderWidth"]
```
nvh: Why aren't we adding these to @isAnimatable?

nvh: Ah, that only gets the value. Maybe an isAnimatableValue and isAnimatableKey?

framer/Animation.coffee
```diff
@@ -292,6 +296,24 @@ class exports.Animation extends BaseClass
...
+
+		for propKey in propKeys
+			keyValueA = if _.isNumber(valueA) then valueA else valueA[propKey]
+			keyValueB = if _.isNumber(valueB) then valueB else valueB[propKey]
+			keyValueA ?= keyValueB
+			keyValueB ?= keyValueA
+			result[propKey] = Utils.mapRange(value, 0, 1, keyValueA, keyValueB)
+
+		if _.uniq(_.values(result)).length is 1
+			result = result[propKeys[0]]
```
nvh: There's a lot happening here that isn't directly clear. Maybe the naming of variables could be better, maybe it needs some comments.

extras/Studio.framer/framer/version
eelco: This is just the template version, so the Studio.framer project can be used with the beta. Does not really belong in this PR, sorry.

framer/Animation.coffee
eelco: Nice one, it’s getting unwieldy indeed.

PR #537, opened 2017-08-21
Feature/multiple shadows

nvh:


framer/Layer.coffee
```diff
@@ -366,12 +362,16 @@ class exports.Layer extends BaseClass
...
-				_.defaults shadow, Framer.Defaults.Shadow
-				shadow.color = new Color(shadow.color)
+			shadows = []
+			for shadow, index in value
+				shadows.push proxiedShadowValue(@, shadow, index)
 			@_setPropertyValue("shadows", shadows)
-			updateShadowStyle(@)
+			@updateShadowStyle()
+
+	updateShadowStyle: ->
```
eelco: This looks like an internal function to me, so `_updateShadowStyle`?

nvh: Well, you might want to call it if you manually change the shadows in the `layer.shadows` array

PR #559, opened 2018-01-16
remove layers from the DOMEventManager

onnlucky:
`Layer._element`s were never removed from the `DOMEventManager` thus leaking a lot of dom nodes and `DOMEventManagerElement` (which are `EventEmitter`s).

Especially relevant for projects that use layers for particles or such.

### DIFF
--- a/framer/DOMEventManager.coffee
+++ b/framer/DOMEventManager.coffee
@@ -3,7 +3,7 @@
 
 Utils = require "./Utils"
 
-EventManagerIdCounter = 0
+EventManagerIdCounter = 1
 
 class DOMEventManagerElement extends EventEmitter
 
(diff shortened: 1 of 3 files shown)


framer/DOMEventManager.coffee
```diff
@@ -42,6 +42,11 @@ class exports.DOMEventManager
 
 		@_elements[element._eventManagerId]
 
+	remove: (element) =>
+		return unless element._eventManagerId
+		delete @_elements[element._eventManagerId]
+		element._eventManagerId = 0
```
nvh: shouldn't we set it to `null` instead of 0 here, to more explicitly specify unsetting it?

onnlucky: `0` is a perfectly good `null` for number types, the advantage is this way the field doesn't switch type, which might help the javascript engine keep the code faster. Also why I made sure counting starts at `1`, before this the first layer using `wrap()` would get two `DOMEventManagerElement`s ...

PR #561, opened 2018-01-23
Fix `copy()` crash on Frames containing frames. Fix https://github.co…

edwinvanrijkom:
…m/motif/company/issues/6457

### DIFF
--- a/framer/SVGBaseLayer.coffee
+++ b/framer/SVGBaseLayer.coffee
@@ -2,9 +2,14 @@
 {Layer, layerProperty} = require "./Layer"
 {Color} = require "./Color"
 
-
 _svgMeasureElement = null
 
+denyCopy = ->
+	if Utils.isFramerStudio()
+		throw new Error("SVGLayer and SVGPath do not support the `copy` method")
+	# else
+	console.warn("SVGLayer and SVGPath do not support the `copy` method")
+	return null
 
 getSVGMeasureElement = (constraints={}) ->
 	if not _svgMeasureElement?
@@ -190,3 +195,7 @@ class exports.SVGBaseLayer extends Layer
 	resetViewbox: =>
 		@_svg.setAttribute("viewBox", "0,0,#{@width},#{@height}")
 		@_svg.removeAttribute("viewBox")
+
+	copy: -> return denyCopy()
+
+	copySingle: -> return denyCopy()
--- a/framer/SVGBaseLayer.coffee
+++ b/framer/SVGBaseLayer.coffee
@@ -1,15 +1,12 @@
 {LayerStyle} = require "./LayerStyle"
 {Layer, layerProperty} = require "./Layer"
 {Color} = require "./Color"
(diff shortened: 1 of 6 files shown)


framer/SVGBaseLayer.coffee
```diff
@@ -2,9 +2,14 @@
...
 {Color} = require "./Color"
 
-
 _svgMeasureElement = null
 
+denyCopy = ->
+	if Utils.isFramerStudio()
+		throw new Error("SVGLayer and SVGPath do not support the `copy` method")
+	# else
+	console.warn("SVGLayer and SVGPath do not support the `copy` method")
```
nvh: This should be SVGGroup and SVGPath, right?

framer/SVGLayer.coffee
```diff
@@ -5,6 +5,13 @@
 {SVGGroup} = require "./SVGGroup"
 {SVGPath} = require "./SVGPath"
 
+denyCopy = ->
+	if Utils.isFramerStudio()
+		throw new Error("SVGLayer does not support the `copy` method")
+	# else
+	console.warn("SVGLayer does not support the `copy` method")
+	return null
```
nvh: This duplication is not really nice, is there a way to wrap it in a shared function?
Also, I would make the wording a bit more accurate, because we _do_ support it if the layer doesn't have any children, right?

framer/SVGBaseLayer.coffee
```diff
@@ -1,15 +1,12 @@
...
 
 _svgMeasureElement = null
 
 denyCopy = ->
-	if Utils.isFramerStudio()
-		throw new Error("SVGLayer and SVGPath do not support the `copy` method")
-	# else
-	console.warn("SVGLayer and SVGPath do not support the `copy` method")
-	return null
+	return Utils.throwInStudioOrWarnInProduction("SVGLayer and SVGPath do not support the `copy` method")
```
nvh: Still, SVGGroup, not SVGLayer, right?

test/tests/UtilsTest.coffee
```diff
@@ -781,3 +781,9 @@ describe "Utils", ->
 				Utils.equal(new Color("red"), 0).should.be.false
 			it "should return false when only right side is a Color", ->
 				Utils.equal(3, new Color("red")).should.be.false
+
+	describe "throw or warn message", ->
+		it "should warn when in production", ->
+			Utils.throwInStudioOrWarnInProduction("test")
+			# not throwing means a succesful test. There's no way to mock being in Framer Studio afaik, that
+			# is, without resorting to adding a mocking lib like sinon.js, which is overkill for just this method.
```
nvh: See how to do this here: https://github.com/koenbok/Framer/blob/master/test/tests/DeviceComponentTest.coffee#L41

edwinvanrijkom: Substitution like that works locally, but not when it's called from another module, afaict.

PR #563, opened 2018-01-30
Feature/stroke api

nvh:


framer/SVGPath.coffee
```diff
@@ -30,6 +30,31 @@ class exports.SVGPath extends SVGBaseLayer
 			@_path = @_svg.querySelector(link)
 		@_length = @_path.getTotalLength()
 
+
+	updateStroke: ->
+		startLength = @strokeStart ? 0
+		endLength = @strokeEnd ? @length
```
onnlucky: I think we need to make sure strokeEnd > length is handled like an angle > 360. Otherwise animations that overshoot have strange effects.

test/tests/SVGPathTest.coffee
```diff
@@ -31,6 +33,119 @@ describe "SVGPath", ->
 			point.x.should.be.closeTo 0, 0.01
 			point.y.should.be.closeTo 50, 0.01
 
+	describe "strokeLength", ->
```
onnlucky: longer than length? less than zero, etc? Test those too?

framer/SVGPath.coffee
nvh: I've tested this, and what we actually want to do is what we do now, limiting strokeEnd to path.length, otherwise you get the weird effects you describe

test/tests/SVGPathTest.coffee
nvh: Correct, there weren't any tests, but the code was correct. I've added them for completeness

PR #564, opened 2018-02-16
autoSize texts should never softwrap

onnlucky:
Synchronizes design behavior and visuals with framerjs runtime. Also fixes
a bug that limiting width to parent.width only works when x = 0.

Setting width will grow in height. Setting width and height will say fixed.

### DIFF
--- a/framer/TextLayer.coffee
+++ b/framer/TextLayer.coffee
@@ -93,14 +93,29 @@ class exports.TextLayer extends Layer
 		super options
 		@__constructor = true
 
-		# Keeps track if the width or height are explicitly set, so we shouldn't update it afterwards
-		if not options.autoSize? and not options.truncate
+		# the goal is:
+		# - autoSize elements should not soft wrap, only hard wrap (based on newlines)
+		# - fixed (given a box) elements should soft wrap
+		# - when the height is not fixed, it should be allowed to grow
+
+		if options.autoSize
+			@autoWidth = true
+			@autoHeight = true
+		else if options.autoSize isnt false and not options.truncate
+			# if not explicitly disabled auto sizing, auto size width/height, unless they were explicitly set
 			if not options.autoWidth?
-				explicitWidth = options.width? or _.isNumber(options?.size) or options?.size?.width? or options?.frame?.width?
+				explicitWidth = options.width? or _.isNumber(options.size) or options.size?.width? or options.frame?.width?
 				@autoWidth = not explicitWidth
 			if not options.autoHeight?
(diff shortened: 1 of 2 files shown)


framer/TextLayer.coffee
```diff
@@ -93,15 +93,27 @@ class exports.TextLayer extends Layer
 		super options
 		@__constructor = true
 
-		# Keeps track if the width or height are explicitly set, so we shouldn't update it afterwards
-		if not options.autoSize? and not options.truncate
+		# the goal is
+		# autoSize elements should not soft wrap, only hard wrap (based on newlines)
+		# fixed (given a box) elements should soft wrap
```
nvh: Some weird newlines here, maybe add `- ` in front of the last two lines and a `:` after the first?

framer/TextLayer.coffee
```diff
@@ -93,15 +93,27 @@ class exports.TextLayer extends Layer
...
+			@autoHeight = true
+		else if options.autoSize isnt false and not options.truncate
+			# if not explicitly disabled auto sizing, auto size width/height, unless they were explicitly set
 			if not options.autoWidth?
-				explicitWidth = options.width? or _.isNumber(options?.size) or options?.size?.width? or options?.frame?.width?
+				explicitWidth = options.width? or _.isNumber(options.size) or options.size?.width? or options.frame?.width?
 				@autoWidth = not explicitWidth
 			if not options.autoHeight?
-				explicitHeight = options.height? or _.isNumber(options?.size) or options?.size?.height? or options?.frame?.height?
+				explicitHeight = options.height? or _.isNumber(options.size) or options.size?.height? or options.frame?.height?
```
nvh: Why dit you remove the optionality of the `options` here?

framer/TextLayer.coffee
```diff
@@ -93,15 +93,27 @@ class exports.TextLayer extends Layer
...
 				@autoHeight = not explicitHeight
 
+		# if constraints from design, autoHeight depends on if the element is allowed to grow in height
+		constraintValues = options.constraintValues
+		if constraintValues
+			@autoHeight = not (_.isNumber(constraintValues.top) and _.isNumber(constraintValues.bottom))
```
nvh: Shouldn't we check for heightFactor here as well? Or am I mistaken

framer/TextLayer.coffee
onnlucky: `options` is never null

framer/TextLayer.coffee
onnlucky: will have to check that ... maybe

